PR #334, opened 2022-10-04
init: update config.mk references from sub systems Makefiles.

delvatt:
Makefiles  within Init system are still using references from  ukama/services/config.mk. 
Maybe they should be updated to use the config inside ukama/systems/config.mk

Signed-off-by: Elvis Attro <attroelvis@gmail.com>


delvatt:
@vthakur7f @markhor  Can we close this if it is already fixed by commit if other PRs?

PR #338, opened 2022-10-10
BFF system workflow updated

salman-codewar:
## Purpose changes

**BFF System**
- Update bff system workflow after it moved to systems


Brackleycassinga (approved):
LGTM

PR #354, opened 2022-11-14
Registry: deactivate users & org members.

### LINKED ISSUE #353: Deactivate users/org-members consistently across users and org sub-systems.
Users are registered throughout the Registry system and can be added individually as members to various organizations, independently.
We need to implement a consistent way to deactivate a member within an org and also,  deactivate a user within the Registry in a way that this deactivates that user in each organization this user is member of.

- [x] Deactivate an Org member
- [x] Deactivate a User

delvatt:
This pul request fixes


delvatt:
This PR is being replaced by #363 which fixes the subsequent issue #353  by commit.

PR #368, opened 2022-12-14
Add gRPC sub-system for managing subscriber operations

Brackleycassinga:
This pull request adds a new gRPC service for managing Subscriber Identity Module (SIM) cards. The service includes methods for adding, deleting, and retrieving information about subscribers.

The added methods are:

`AddSubscriber:`

```
 Adds a new subscriber to the database with the provided information. Returns a unique ID for the subscriber.
```

`DeleteSubscriber :` 

```
Deletes the subscriber with the provided ID from the database.
```

`GetSubscriber:`

```
 Retrieves the information for the subscriber with the provided ID.
 ```
 
Overall, this pull request provides a robust and efficient way to manage subscribers and SIM cards using gRPC.

### CODE REVIEW

File: systems/subscriber/api-gateway/pb/simManager.proto [hunk 1]
```diff
@@ -0,0 +1,140 @@
+syntax = "proto3";
```
delvatt, line 1: We should not be keeping other services proto files as this might leads to proto duplication and will make it harder to keep these files in sync or to maintain. Instead, depend directly on the other system gen package.

systems/subscriber/api-gateway/pb/simPool.proto
```diff
@@ -0,0 +1,75 @@
+syntax = "proto3";
```
delvatt, line 1: We should not be keeping other services proto files as this might leads to proto duplication and will make it harder to keep these files in sync or to maintain. Instead, depend directly on the other system gen package.

systems/subscriber/api-gateway/pb/subscriber.proto
```diff
@@ -0,0 +1,91 @@
+syntax = "proto3";
```
delvatt, line 1: We should not be keeping other services proto files as this might leads to proto duplication and will make it harder to keep these files in sync or to maintain. Instead, depend directly on the other system gen package.

systems/subscriber/subscriber-registry/pb/subscriber.proto
```diff
@@ -0,0 +1,190 @@
...
+// GetByNetworkResponse message is used to get all subscribers by network id
+message GetByNetworkResponse{
+// Repeated field of Subscriber message
+repeated Subscriber subscribers = 1;
+}
+
+// DeleteSubscriberResponse message is used to delete a subscriber
+message DeleteSubscriberResponse{
+// subscriberID field is used to specify the subscriber id for deletion
+string subscriberID = 1;
```
delvatt, line 63: Why do we return the ID of the deleted subscriber, especially if that ID is no more valid (because the subscriber is deleted) ? And also, we already have this information as this is the ID that we pass to the delete request.

systems/subscriber/subscriber-registry/pb/subscriber.proto
```diff
@@ -0,0 +1,190 @@
...
+    // idSerial of the subscriber
+    string idSerial =6;
+    // network ID of the subscriber, must be a UUID and not empty
+    string networkID = 7 [(validator.field) = {uuid_ver: 0, string_not_empty: true}, json_name = "network_id"];
+    // proof of identification of the subscriber, must not be empty and have a length greater than 1
+    string proofOfIdentification =8 [(validator.field) = { string_not_empty: true, length_gt: 1 }];
+    // date of birth of the subscriber
+    google.protobuf.Timestamp dateOfBirth = 9;
+    // gender of the subscriber, must not be empty and have a length greater than 1
+    string gender =10 [(validator.field) = { string_not_empty: true, length_gt: 1 }];
```
delvatt, line 99: Do we really need the gender information? Is there any specific thing that our system needs to do depending on that information?

systems/subscriber/subscriber-registry/pb/subscriber.proto
```diff
@@ -0,0 +1,190 @@
...
+    // org ID of the subscriber, must be a UUID and not empty
+    string orgID = 11 [(validator.field) = {uuid_ver: 0, string_not_empty: true}, json_name = "org_id"];
```
delvatt, line 101: If we are keeping both the networkID and the orgID, then we need to make sure that these two information never mismatch.

systems/subscriber/subscriber-registry/pb/subscriber.proto
```diff
@@ -0,0 +1,190 @@
...
+    }
+message AddSubscriberResponse{
+    Subscriber Subscriber = 1;
+
+}
+message Subscriber{
+    string orgID = 1 [(validator.field) = {uuid_ver: 0, string_not_empty: true}, json_name = "org_id"];
+    string firstName  = 2;
+    string lastName = 3;
+    string subscriberID = 4 [json_name = "subscriber_id"];
```
delvatt, line 139: should be enforcing uuid_version and non emptiness.

systems/subscriber/subscriber-registry/pb/subscriber.proto
```diff
@@ -0,0 +1,190 @@
...
+    string networkID = 5;
```
delvatt, line 140: should be enforcing uuid_version and non emptiness for networkID

systems/subscriber/subscriber-registry/pkg/db/model.go
```diff
@@ -0,0 +1,25 @@
+package db
+
+import (
+	"time"
+
+	"github.com/gofrs/uuid"
+)
+
+type Subscriber struct {
+	SubscriberID          uuid.UUID `gorm:"type:uuid;index"`
```
delvatt, line 10: Does gorm `index` tag enforce field uniqueness? Not so sure of that. Please, confirm if so. If not, please explicitly enforce the `primary key` constraint for subscirberID field.

systems/subscriber/subscriber-registry/pkg/db/subscriber_repo.go
```diff
@@ -0,0 +1,84 @@
+package db
+
+import (
+	"github.com/gofrs/uuid"
```
delvatt, line 4: common/uuid

systems/subscriber/subscriber-registry/pkg/db/subscriber_repo.go
```diff
@@ -0,0 +1,84 @@
...
+	Update(subscriberID uuid.UUID, sub Subscriber) (*Subscriber, error)
+	GetByNetwork(networkID uuid.UUID) ([]Subscriber, error)
+	ListSubscribers() ([]Subscriber, error)
+}
+
+type subscriberRepo struct {
+	Db sql.Db
+}
+
+func NewSubscriberRepo(db sql.Db) *subscriberRepo {
```
delvatt, line 22: Please, return the SubscriberRepo interface instead.

systems/subscriber/subscriber-registry/pkg/db/subscriber_repo.go
```diff
@@ -0,0 +1,84 @@
...
+	if result.Error != nil {
+		return nil, result.Error
+	}
+	return subscribers, nil
+}
+
+func (s *subscriberRepo) Get(subscriberId uuid.UUID) (*Subscriber, error) {
+	var subscriber Subscriber
+	
+	err := s.Db.GetGormDb().Where("subscriber_id = ?", subscriberId).FirstOrCreate(&subscriber).Error
```
delvatt, line 45: Should the call to get a given subscriber, create a new record if that subscriber does not exist?

systems/subscriber/subscriber-registry/pkg/db/subscriber_repo_test.go
```diff
@@ -0,0 +1,124 @@
+package db
```
delvatt, line 1: we can add more tests for the other repo methods.

systems/subscriber/subscriber-registry/pkg/server/event.go
```diff
@@ -0,0 +1,49 @@
...
+func NewSubscriberEventServer(subscriberRepo db.SubscriberRepo) *SubcriberRegistryEventServer {
+	return &SubcriberRegistryEventServer{
+		subscriberRepo: subscriberRepo,
+	}
+}
+
+func (l *SubcriberRegistryEventServer) EventNotification(ctx context.Context, e *epb.Event) (*epb.EventResponse, error) {
+	log.Infof("Received a message with Routing key %s and Message %+v", e.RoutingKey, e.Msg)
+	switch e.RoutingKey {
+	case "event.cloud.simManager.sim.allocation":
```
delvatt, line 28: make sure the listened route matches the publishers'

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
+package server
+
+import (
+	"context"
+
+	"github.com/gofrs/uuid"
```
delvatt, line 6: common/uuid

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+	return &SubcriberServer{subscriberRepo: subscriberRepo,
+		msgbus:         msgBus,
+		subscriberRoutingKey: msgbus.NewRoutingKeyBuilder().SetCloudSource().SetContainer(pkg.ServiceName)}
+}
+
+
+
+func (s *SubcriberServer) Add(ctx context.Context, req *pb.AddSubscriberRequest) (*pb.AddSubscriberResponse, error) {
+	logrus.Infof("Adding subscriber: %v", req)
+	networkID_uuid := uuid.FromStringOrNil(req.GetNetworkID())
```
delvatt, line 38: 1) Use camelCase notation instead of snake_case.
2) Use either networkID or networkUUID, but not both.
3) What do we do when networkID is nil?

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+	orgID_uuid := uuid.FromStringOrNil(req.GetOrgID())
```
delvatt, line 39: 1) Use camelCase notation instead of snake_case.
2) Use either orgID or orgUUID, but not both.
3) What do we do when orgID is nil?

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+
+	subscriberID_uuid, err := uuid.NewV4()
```
delvatt, line 41: same here. no snake_case.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+			Address:               req.GetAddress(),
+			ProofOfIdentification: req.GetProofOfIdentification(),
+			DateOfBirth:           birthday.GoString(),
+			IdSerial:              req.GetIdSerial()},
+	}, nil
+
+}
+
+func (s *SubcriberServer) Delete(ctx context.Context, req *pb.DeleteSubscriberRequest) (*pb.DeleteSubscriberResponse, error) {
+	subscriberID := req.GetSubscriberID()
```
delvatt, line 93: doing later on `	subscriberIDUUID, err := uuid.FromString(subscriberID)`  makes this line useless.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+	if subscriberID == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "subscriberID must not be empty")
+	}
+
+	if err := ctx.Err(); err != nil {
```
delvatt, line 98: why are we doing this?

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, err
+	}
+	subscriberIDUUID, err := uuid.FromString(subscriberID)
```
delvatt, line 101: 1) subscriberID or subscriberUUID
2) We never swallow errors in Go. Please always handle any resulting error from any given operation.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+			ProofOfIdentification: req.GetProofOfIdentification(),
+			DateOfBirth:           birthday.GoString(),
+			IdSerial:              req.GetIdSerial()},
+	}, nil
+
+}
+
+func (s *SubcriberServer) Delete(ctx context.Context, req *pb.DeleteSubscriberRequest) (*pb.DeleteSubscriberResponse, error) {
+	subscriberID := req.GetSubscriberID()
+	if subscriberID == "" {
```
delvatt, line 94: Doing later on `	subscriberIDUUID, err := uuid.FromString(subscriberID)`  makes this check useless.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+	if err != nil {
+		logrus.WithError(err).Error("error while deleting subscriber")
+		return nil, grpc.SqlErrorToGrpc(err, "subscriber")
+	}
+	route := s.subscriberRoutingKey.SetAction("delete").SetObject("subscriber").MustBuild()
+	err = s.msgbus.PublishRequest(route, req)
+	if err != nil {
+		logrus.Errorf("Failed to publish message %+v with key %+v. Errors %s", req, route, err.Error())
+	}
+	if err != nil {
```
delvatt, line 114: We are checking the error twice.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+	return &pb.DeleteSubscriberResponse{
+		SubscriberID: subscriberID,
+	}, nil
+}
+
+func (s *SubcriberServer) Get(ctx context.Context, req *pb.GetSubscriberRequest) (*pb.GetSubscriberResponse, error) {
+
+	subscriberID := req.GetSubscriberID()
+	if subscriberID == "" {
```
delvatt, line 125: Again, no need to do this check.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, status.Errorf(codes.InvalidArgument, "subscriberID must not be empty")
+	}
+
+	if err := ctx.Err(); err != nil {
```
delvatt, line 129: What is that for?

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+
+	subscriberList := &pb.ListSubscribersResponse{
+		Subscribers: dbsubscriberToPbSubscribers(subscribers),
+	}
+
+	return subscriberList, nil
+}
+func (s *SubcriberServer) GetByNetwork(ctx context.Context, req *pb.GetByNetworkRequest) (*pb.GetByNetworkResponse, error) {
+	networkID := req.GetNetworkID()
+	if networkID == "" {
```
delvatt, line 171: same as above

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, status.Errorf(codes.InvalidArgument, "networkID must not be empty")
+	}
+
+	if err := ctx.Err(); err != nil {
```
delvatt, line 175: same as above

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, err
+	}
+
+	logrus.Infof("Get subscribers by network: %v ", networkID)
+	networkIDUUID, err := uuid.FromString(networkID)
```
delvatt, line 180: networkID or networkUUID

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+
+	return subscriberList, nil
+}
+func (s *SubcriberServer) Update(ctx context.Context, req *pb.UpdateSubscriberRequest) (*pb.UpdateSubscriberResponse, error) {
+	subscriberIdReq := req.GetSubscriberID()
+	if subscriberIdReq == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "subscriberID must not be empty")
+	}
+
+	if err := ctx.Err(); err != nil {
```
delvatt, line 203: same as above

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, err
+	}
+	subscribeIDUUID, err := uuid.FromString(subscriberIdReq)
```
delvatt, line 206: 1) subscriberIDUUID does not make sense.
2) err is being swallowed.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, grpc.SqlErrorToGrpc(err, "subscribers")
+	}
+
+	subscriberList := &pb.ListSubscribersResponse{
+		Subscribers: dbsubscriberToPbSubscribers(subscribers),
+	}
+
+	return subscriberList, nil
+}
+func (s *SubcriberServer) GetByNetwork(ctx context.Context, req *pb.GetByNetworkRequest) (*pb.GetByNetworkResponse, error) {
```
delvatt, line 169: This RPC is missing the client call to sim-manager to get the list of associated sims.

systems/subscriber/subscriber-registry/pkg/server/subscriber_test.go
```diff
@@ -0,0 +1,170 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/gofrs/uuid"
```
delvatt, line 9: common/uuid

systems/subscriber/subscriber-registry/pkg/db/model.go
Brackleycassinga: You're right.I have changed it to be unique and primaryKey .index only help in filtering data.

systems/subscriber/subscriber-registry/pkg/db/model.go
```diff
@@ -0,0 +1,25 @@
+package db
+
+import (
+	"time"
+
+	"github.com/gofrs/uuid"
```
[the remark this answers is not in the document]
Brackleycassinga: This have been changed to common.

systems/subscriber/subscriber-registry/pkg/db/subscriber_repo.go
Brackleycassinga: Same here

systems/subscriber/subscriber-registry/pkg/db/subscriber_repo.go
Brackleycassinga: Fixed.

systems/subscriber/subscriber-registry/pkg/db/subscriber_repo.go
Brackleycassinga: I have changed it to call .`First` instead .

systems/subscriber/subscriber-registry/pkg/db/subscriber_repo_test.go
Brackleycassinga: Sure

systems/subscriber/subscriber-registry/pkg/server/event.go
Brackleycassinga: I need to  return req for `Sim allocation ` method here but I don't want to merge simManager on my branch now.I want this branch to be merged to fix those small issues.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: Addressed.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: Addressed.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: We check if there's an error in the context because it ensures that the request processing is not continued if the context has already been canceled or if the deadline has already been exceeded. This helps to prevent any unnecessary resource utilization or processing in case the client has already lost interest in the request.

I read about it here Lhttps://tutorialedge.net/golang/go-grpc-context-cancellation-tutorial/

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: Addressed.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: I have removed the first check and I handle `uuid.FromString(subscriberID) `error.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: Thanks.Legacy has been removed.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: This to check the context error .I read about seems like a good practice.I have  added a more in depth desc in the first comment.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: Thanks.All legacy code have been removed.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: I have added comment earlier in this thread.

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
Brackleycassinga: There is conflict when I try to integrate your branch in mine.Once all sub-systems are merged.What do you think ?

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
delvatt: Thanks. You should share the article on our slack, so that we can discuss the practice all together.

systems/subscriber/subscriber-registry/go.mod
```diff
@@ -0,0 +1,86 @@
+module github.com/ukama/ukama/systems/subscriber/subscriber-registry
+
+go 1.18
+
+replace github.com/ukama/ukama/systems/common => ../../common
+
+replace github.com/ukama/ukama/systems/init/msgClient => ../../init/msgClient
+
+require (
+	github.com/gofrs/uuid v4.3.1+incompatible
```
delvatt, line 10: should not be present anymore

systems/subscriber/subscriber-registry/pkg/server/subscriber.go
```diff
@@ -0,0 +1,260 @@
...
+	if orgID == uuid.Nil {
+		return nil, status.Errorf(codes.InvalidArgument, "OrgID must not be empty")
+	}
+	subscriberID := uuid.NewV4()
+	timestamp := &timestamppb.Timestamp{
+		Seconds: req.DateOfBirth.GetSeconds(),
+		Nanos:   req.DateOfBirth.GetNanos(),
+	}
+
+	birthday := timestamp.AsTime()
```
delvatt, line 49: You should try `req.DateOfBirth.AsTime()`

delvatt (commented):
Good. Thanks for addressing  all the remarks. Just the  remote client to implement and we are good to go!

systems/subscriber/subscriber-registry/go.mod
Brackleycassinga: Package removed.

PR #380, opened 2023-01-20
Msg client

vthakur7f:
Implememnts a first version of the client service to RabbitMQ aka msgBus.
Services need to register to this client service and Implement the GRPC Event server, to listen for incoming events.
Apart from listening and publishing , this service also implement the custom keep alive grpc message to all registered services in the system. This could be useful later to check service state in system.


Brackleycassinga (approved):
LGTM

systems/common/config/config.go
```diff
@@ -187,3 +197,14 @@ func DefaultForwardConfig() Forward {
 		Path: "/",
 	}
 }
+
+func LoadServiceHostConfig(name string) *Service {
+	svcHost :="_SERVICE_HOST"
+	svcPort :="_SERVICE_PORT"
```
Brackleycassinga, line 203: I think you should check whether the enviroment variables exist before using them.
```
if host == "" {
		logrus.Errorf("environment variable %s not set", strings.ToUpper(name+svcHost))
		return nil, errors.New("environment variable not set")
	}
```

systems/common/msgBusServiceClient/msgbus_service_client.go
```diff
@@ -0,0 +1,151 @@
...
+
+}
+
+func (m *msgBusServiceClient) Register() error {
+	logrus.Debugf("Registering %s service instance %s to MessageBusClient at %s.", m.service, m.instanceId, m.msgClientURI)
+	ctx, cancel := context.WithTimeout(context.Background(), m.timeout)
+	defer cancel()
+
+	resp, err := m.client.RegisterService(ctx, &pb.RegisterServiceReq{
+		ServiceName: m.service,
```
Brackleycassinga, line 80: I think you need to verify that all fields of m *msgBusServiceClient have values before use them in the req .and check the REGISTRATION_STATUS  to handle different states.what happened when a service is already registered ?I think we should show some type of logs and not ovewrite the old instance.

systems/common/msgBusServiceClient/msgbus_service_client.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+func (m *msgBusServiceClient) Start() error {
+	logrus.Debugf("Starting MessageClientRoutine for %s service instance %s Routine ID %s.", m.service, m.instanceId, m.uuid)
+	ctx, cancel := context.WithTimeout(context.Background(), m.timeout)
+	defer cancel()
+
+	_, err := m.client.StartMsgBusHandler(ctx, &pb.StartMsgBusHandlerReq{
+		ServiceUuid: m.uuid})
```
Brackleycassinga, line 108: Please check if `m.uuid() `has been generated successfully before you can assign it to the req.

systems/common/msgBusServiceClient/msgbus_service_client.go
```diff
@@ -0,0 +1,151 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), m.timeout)
+	defer cancel()
+
+	anyMsg, err := anypb.New(msg)
+	if err != nil {
+		return err
+	}
+
+	_, err = m.client.PublishMsg(ctx, &pb.PublishMsgRequest{
+		ServiceUuid: m.uuid,
```
Brackleycassinga, line 142: Same here m.uuid is assigned before check if it has been successfully generated .

systems/common/msgBusServiceClient/msgbus_service_client.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+func (m *msgBusServiceClient) Stop() error {
+	logrus.Debugf("Stopping MessageClientRoutine for %s service instance %s Routine ID %s.", m.service, m.instanceId, m.uuid)
+	ctx, cancel := context.WithTimeout(context.Background(), m.timeout)
+	defer cancel()
+
+	_, err := m.client.StopMsgBusHandler(ctx, &pb.StopMsgBusHandlerReq{
+		ServiceUuid: m.uuid,
```
Brackleycassinga, line 122: Same here m.uuid is assigned before check if it has been successfully generated .

systems/common/msgbus/qpub.go
```diff
@@ -55,6 +57,27 @@ func (q *qPub) Publish(payload any, routingKey string) error {
 	return nil
 }
 
+func (q *qPub) PublishProto(payload proto.Message, routingKey string) error {
+
+	b, err := proto.Marshal(payload)
+	if err != nil {
+		return err
```
Brackleycassinga, line 64: Can you use logrus to log the errors when marshal does not and when function fail to publish.I think it will useful in debugging.

systems/common/msgbus/routing_key.go
```diff
@@ -145,3 +140,30 @@ func (r RoutingKeyBuilder) MustBuild() string {
 	}
 	return res
 }
+
+func Parse(s string) (RoutingKey, error) {
+
+	parts := strings.Split(s, ".")
+	if len(parts) != 5 {
+		return "", fmt.Errorf("invalid route %s", s)
```
Brackleycassinga, line 148: Please use logrus to output errors instead.

systems/init/lookup/cmd/lookup/main.go
```diff
@@ -16,13 +16,14 @@ import (
 	log "github.com/sirupsen/logrus"
```
Brackleycassinga, line 16: logrus import is duplicated.

systems/init/lookup/cmd/lookup/main.go
```diff
@@ -71,23 +67,51 @@ func initConfig() {
...
 }
 
 func runGrpcServer(d sql.Db) {
+
 	instanceId := os.Getenv("POD_NAME")
+	if instanceId == "" {
+		/* used on local machines */
+		inst, err := uuid.NewV4()
+		if err != nil {
+			log.Fatalf("Failed to genrate instanceId. Error %s", err.Error())
```
Brackleycassinga, line 82: Use logrus instead as we have used it everywhere in our code base.

systems/init/lookup/cmd/lookup/main.go
```diff
@@ -71,23 +67,51 @@ func initConfig() {
...
 
+	go msgBusListener(mbClient)
+
 	grpcServer.StartServer()
 }
+
+func msgBusListener(m mb.MsgBusServiceClient) {
+
+	if err := m.Register(); err != nil {
+		log.Fatalf("Failed to register to Message Client Service. Error %s", err.Error())
```
Brackleycassinga, line 111: Use logrus instead

systems/init/lookup/internal/server/event.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+}
+
+func (l *LookupEventServer) EventNotification(ctx context.Context, e *epb.Event) (*epb.EventResponse, error) {
+	log.Infof("Received a message with Routing key %s and Message %+v", e.RoutingKey, e.Msg)
+	switch e.RoutingKey {
+	case "event.cloud.lookup.organization.create":
+		msg, err := unmarshalLookupOrganizationCreate(e.Msg)
+		if err != nil {
+			return nil, err
```
Brackleycassinga, line 35: use logrus to log all errors

systems/init/lookup/internal/server/event.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+		err = handleEventCloudLookupOrgCreate(e.RoutingKey, msg)
+		if err != nil {
+			return nil, err
+		}
```
Brackleycassinga, line 41: Use logrus

systems/init/lookup/internal/server/event.go
```diff
@@ -0,0 +1,62 @@
...
+		log.Errorf("No handler routing key %s", e.RoutingKey)
+	}
+
+	return &epb.EventResponse{}, nil
+}
+
+func unmarshalLookupOrganizationCreate(msg *anypb.Any) (*pb.AddOrgRequest, error) {
+	p := &pb.AddOrgRequest{}
+	err := anypb.UnmarshalTo(msg, p, proto.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true})
+	if err != nil {
```
Brackleycassinga, line 52: I think you need to use logrus again here to logs all errors for tracing and debugging.

systems/init/msgClient/internal/config.go
```diff
@@ -0,0 +1,31 @@
...
+
+type Config struct {
+	uconf.BaseConfig `mapstructure:",squash"`
+	DB               *uconf.Database   `default:"{}"`
+	Grpc             *uconf.Grpc       `default:"{}"`
+	Queue            *uconf.Queue      `default:"{}"`
+	Metrics          *uconf.Metrics    `default:"{}"`
+	Timeout          time.Duration     `default:"3s"`
+	HeathCheck       HeathCheckRoutine `default:"{}"`
+	System           string            `default:"init"`
```
Brackleycassinga, line 17: Can we make var `System` a mapstructure:"system"  in order to make the System variable configurable ?

systems/init/msgClient/internal/queue/publisher.go
```diff
@@ -0,0 +1,76 @@
...
+	return qp, nil
+}
+
+func (p *QueuePublisher) Publish(key string, payload proto.Message) error {
+
+	err := make(chan error, 1)
+	go func(err chan error) {
+		e := p.pub.PublishProto(payload, key)
+		if e != nil {
+			log.Errorf("Failed to publish message. Error %s", e.Error())
```
Brackleycassinga, line 46: Use logrus instead here too.

Brackleycassinga (changes requested):
Great work Vishal 👍🏾 
I have added some comments.

systems/common/msgBusServiceClient/msgbus_service_client.go
vthakur7f: Service will register once in it's life time. I don see any use case where service woul re-register itself. For upgardes or crash  , service would anyway request for registration again on startup. In that all new parameters will be taken by msgClient sdervice but the UUID provided with the previous existing one.  let me know if I missed anything or  you have any use case where we would need it.

vthakur7f: Also check is avaibale for registration afetr error handling

systems/common/msgBusServiceClient/msgbus_service_client.go
vthakur7f: Updated the registration response handling if not registered it will produce error.

systems/common/msgBusServiceClient/msgbus_service_client.go
vthakur7f: Service will not start without this UUID

systems/common/msgbus/routing_key.go
vthakur7f: returning error here.

systems/init/msgClient/internal/config.go
vthakur7f: just define System in env that should be good enough

systems/init/lookup/internal/server/event.go
vthakur7f: I think logging imp error is needed but not at all levels. Like unmarshalLookupOrganizationCreate has already logged it. So I think it should be good.

systems/common/config/config.go
vthakur7f: I used os.LookuoEnv. It will set the env if present otherwise default will be used.

Brackleycassinga (approved):
LGTM 👍🏾

markhor (approved):

systems/init/msgClient/pb/msgClient.proto
```diff
@@ -0,0 +1,88 @@
...
+    rpc UnregisterService(UnregisterServiceReq) returns (UnregisterServiceResp);
+    /// Call this rpc to publisg events
+    rpc PublishMsg(PublishMsgRequest) returns (PublishMsgResponse);
+
+}
+
+/// Registration status enums
+enum REGISTRAION_STATUS {
+    REGISTERED = 0; /// System registered status
+    ALLREADY_REGISTERED = 1; /// System already registered status
```
delvatt, line 35: I guess you meant "ALREADY_REGISTERED". Also, isn't an "ALREADY_REGISTERED" status a "REGISTERED" status?

vthakur7f: yeah. Idea was to get ALREADY_REGISTERED when service was previously succesfully registered, But its not required anymore. So removed.

systems/init/lookup/cmd/lookup/main.go
vthakur7f: log is uased as alias for logrus. For this and all other comments regarding log.

vthakur7f:
msgClient service a interface to message bus (RabbitMQ)

delvatt (approved):
Good job so far!

PR #384, opened 2023-01-24
Subscriber/Test agent

### LINKED ISSUE #382: Implement Test Agent for the Subscriber system.
The Test Agent is a dummy implementation of the Agent interface that is used by the Sim Manager to manage sim cards and perform more specialized sim operations within the subscriber system. It implements real sims operations in a dummy fashion, only for testing purposes.

delvatt:
fixes #382


systems/subscriber/test-agent/cmd/server/main.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			logrus.Infof("Config:\n%s", string(b))
+		}
+	}
+
+	pkg.IsDebugMode = svcConf.DebugMode
+}
+
+func runGrpcServer() {
+	testAgentServer := server.NewTestAgentServer(storage.NewMemStorage())
```
vthakur7f, line 65: Are we using etcd anywhere? I can't find it used anywhere.

systems/subscriber/test-agent/pb/test_agent.proto
```diff
@@ -0,0 +1,54 @@
+syntax = "proto3";
+
+option go_package = "pb/gen";
+
+import "validator.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
+
+
+package ukama.test_agent.v1;
```
vthakur7f, line 10: I think we should update packages to ukama.subscriber.test_agent.v1

systems/subscriber/test-agent/pkg/server/test_agent.go
```diff
@@ -0,0 +1,142 @@
...
+	pb.UnimplementedTestAgentServiceServer
+}
+
+func NewTestAgentServer(storage storage.Storage) *TestAgentServer {
+	return &TestAgentServer{
+		storage: storage,
+	}
+}
+
+func (s *TestAgentServer) GetSimInfo(ctx context.Context, req *pb.GetSimInfoRequest) (*pb.GetSimInfoResponse, error) {
```
vthakur7f, line 27: Does this mean before Acitivating Sim we need to call GetSimInfo to create a sim ?

vthakur7f (commented):
Nice work..!! Just few questions otherwise it's good.

vthakur7f (approved):
Based on discussion I had with @delvatt. Small updates will be done.

systems/subscriber/test-agent/cmd/server/main.go
delvatt: No. Not for now. Therefore removed.

systems/subscriber/test-agent/pb/test_agent.proto
delvatt: Okay.  Until now, our proto packages paths were set to match the following pattern: `ukama.service_name.version`. I understant that now, we are going to update all of them to match the following pattern: `ukama.system_name.service_name.version`

systems/subscriber/test-agent/pkg/server/test_agent.go
delvatt: No. If the sim does already exist, then we activate it by  making an `ActivateSim` RPC call if we are directly talking to the agent  or a `ToggleSimStatus` RPC if we talking to the agent via Sim Manager. The `GetSim`  (previously `GetSimInfo` ) will only create a new sim info for the first time if the sim does not exist.

vthakur7f:
Any open items shall be addressed by opening issues.

delvatt:
And any new Pull request should be made against the `subscriber-system` branch

Issue #385, opened 2023-01-25
MsgClient: Service config error


vthakur7f:
**Describe the bug**
Host value is overwritten by port.

**Expected behavior**
Host ans Port value should eb set properly

**Actual Behavior**
Host value is overwritten by port

**To Reproduce**
1. Start the service which use Service config.
2. Print this config in main you.

**Additional context**
Found during integration test.

PR #386, opened 2023-01-25
Fixed service config Host value.

vthakur7f:
Signed-off-by: vthakur7f <vthakur7@hotmail.com>
#385 : issue with service config fixed.


Brackleycassinga (approved):
LGTM

PR #388, opened 2023-01-26
common/uuid: add uuid wrapper package around github.com/satori/go.uuid.

### LINKED ISSUE #387: Wrap UUID package provider inside common utilities.
In order to avoid using different UUID package providers across all the code base, we need to pick one package and wrap it inside our own within our common utilities.

delvatt:
This PR fixes #387 
Signed-off-by: Elvis Attro <attroelvis@gmail.com>

### DIFF
--- a/systems/common/uuid/uuid.go
+++ b/systems/common/uuid/uuid.go
@@ -12,6 +12,9 @@ import (
 	suuid "github.com/satori/go.uuid"
 )
 
+// Nil represents the nil UUID
+var Nil = UUID{}
+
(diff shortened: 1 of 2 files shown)


vthakur7f (approved):
Looks good. We could have added these function on need basis. 
Only missing is part is test cases. If you could add those before merging it would be great.

delvatt:
> Looks good. We could have added these function on need basis. Only missing is part is test cases, If you could add those before merging it would be great.

@vthakur7f  Just added some unit test cases.  You may want to have a look!

PR #395, opened 2023-02-03
Ukama agent api gateway

vthakur7f:
API Gateway for Ukama Agent.


systems/ukama-agent/api-gateway/pkg/global.go
```diff
@@ -0,0 +1,5 @@
+package pkg
+
+const ServiceName = "api-gateway"
+const SystemName = "ukama-gateway"
```
delvatt, line 4: @vthakur7f do you think we should standardize both service and system name? Because right now we have all type of cases, eg. `ukama-gateway`, `ukama_gateway` and `ukamaGateway`

delvatt (approved):
Lgtm!

systems/ukama-agent/api-gateway/pkg/global.go
vthakur7f: Yeah good point. I myself was wondering about it and when I went through various naming convevtion we are using. I found -(hypen) one more commonly used in our code base. So I went with it. I think it's a kind of thing you pick one and follow.We can do same. Let me know which one we want to follow.

PR #400, opened 2023-02-06
 Registry Updates

### LINKED ISSUE #399: Subscriber/Registry
**Makefile:**
I think we should keep the usual `test` target for running vanilla unit tests quickly and define a new target, like `cover` for running the tests with a coverage profile.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/makefile#L10-L11

You are not using any internal package.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/makefile#L20

**pb:**
I think the service name should be `registry` and not `subscriber-registry` (or `subscriber_registry`). We are already in the `subscriber` systerm and naming one of its internal service _subscriber_registry_ feels redundant. 
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pb/subscriber.proto#L5

**pkg:**
I think the service name should be `registry` and not `subscriber-registry` (or `subscriber_registry`). We are already in the `subscriber` systerm and naming one of its internal service _subscriber_registry_ feels redundant. 
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/global.go#L3

Why is the default host for sim-manager called org?
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/config.go#L17

Use the types defined inside common/rest instead.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/rest/restClient.go#L10-L17

Unnecessary definition of already defined & imported  types.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/client/network.go#L12-L19

Does a `NewNetworkClient`  give us back a new `Network` or a new `NetworkClient` ?
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/client/network.go#L28-L41

The `else` block is unnecessary because of the return statement on line  66. Remove the `else` block and indent back line 68 (`logrus.Infof("Network Info: %+v", network)`)  to the same level of the `if` statements.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/client/network.go#L64-L71

A primary key is already unique by definition.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/db/model.go#L10

Duplicate entry due to unresolved merge conflict.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/db/model.go#L13-L19

These are not DB models repository tests (we should not have any reference to protocol buffers). These are just gRPC server handlers tests. 
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/db/subscriber_repo_test.go#L1-L157

**Err codes & status:**
Not necessarily empty. Even a non empty, but malformed uuid string will generate a parse error and return a Nil UUID result.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L48

Same as above.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L51

Please, return proper grpc error codes & status
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L54-L57

**Nil pointer:**
The following might causes invalid memory  access or nil pointer de-reference. Let's make sure we handle that! 
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L153-L157

https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L195-L203

https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L241-L248

Why do we `log.Fatal` here? 
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L203-L206

Same here,
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L248-L255

In golang, `error` is a keyword that represents an interface. Instead, use`err` in order to avoid any confusion. 
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L218

Where are we calling this function?
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L238-L257

**Time complexity:**
This surely works, but will definitely takes **O(number_of_subscribers*number_of_sims)** time.  Maybe you should consider using a map in order to reduce the time complexity to **O(number_of_subscribers) + O(number_of_sims)**. 
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L159-L172

Same as above.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber.go#L259-L272

**Test and mocks:**
You are defining and implementing your own mock object. This is good when your module/package does not have any dependency on third party mocking libraries. But we are already using those in this project and you already have all the mocks interfaces generated on your **/mocks**  folder, so that you will no have to write your own. You should consider using those interfaces instead.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/pkg/server/subscriber_test.go#L122-L172

**cmd/main:**
Package already imported as `client`
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/cmd/server/main.go#L19

this is not necessary
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/cmd/server/main.go#L36

**Integration:**
The `if` block seems unnecessary to me.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/test/integration/susbcriber_test.go#L69-L71

Same as above.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/test/integration/susbcriber_test.go#L78-L80

I am not sure if this is a good idea to assert exact equality on an error description. Errors can be wrapped inside other errors and in that case, assert.Equal on the error message will fail, even if the error was the correct one. Maybe you should consider using pattern matching instead of exact equality. Look up on the documentation of your assertion library.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/test/integration/susbcriber_test.go#L86

Same as above.
https://github.com/ukama/ukama/blob/6d6757f53b75b65092071702ad73eb5e6537a99e/systems/subscriber/subscriber-registry/test/integration/susbcriber_test.go#L93

Brackleycassinga:
This pull request updates the Subscriber Registry with the following changes:

The file test target has been updated (commit 3b2a487).
The repository has been renamed from "subscriber-registry" to "registry" (commit 3324a81 and 91a2de4).
The primary key unique tag has been removed (commit d0522a5).
The repository test has been updated and a delete repository test has been added (commit cf738b0 and 277fe21).
The server test has been updated (commit f9eab44).
All server and repo tests have been updated to use existing mocks.
Fixes #399


delvatt:
Hey @Brackleycassinga   `Merge branch 'main' into subscriber-registry-review` this is exactly the kind of stuff that we do not want to do anymore, because now,  `subscriber-registry-review` will have all the files and modifications that were made in main (which will include any new file which is in `main` but outside of `subscriber-system` and the reviewer does not need all of those). Also, you can no longer say that `subscriber-system` is the natural base branch of `subsriber-registry-review`.

systems/subscriber/registry/cmd/server/main.go
```diff
@@ -77,17 +73,17 @@ func runGrpcServer(gormdb sql.Db) {
 	}
 	network, err := client.NewNetworkClient(serviceConfig.NetworkHost, pkg.IsDebugMode)
 	if err != nil {
-		log.Fatalf("Network Client initilization failed. Error: %v", err)
+		logrus.Error("Network Client initilization failed. Error: %v", err.Error())
```
delvatt, line 76: Here is exactly the time when we need to `log.Fatal`.

systems/subscriber/registry/pkg/server/subscriber.go
```diff
@@ -26,37 +23,38 @@ type SubcriberServer struct {
...
+		return nil, status.Errorf(codes.InvalidArgument,
+			"invalid format of network uuid. Error %s", nUuidErr.Error())
 	}
-	if orgID == uuid.Nil {
-		return nil, status.Errorf(codes.InvalidArgument, "OrgID must not be empty")
+	orgID, orgUuidErr := uuid.FromString(req.GetOrgID())
+
+	if nUuidErr != nil {
+		return nil, status.Errorf(codes.InvalidArgument,
+			"invalid format of org uuid. Error %s", orgUuidErr.Error())
```
delvatt, line 50: `nUUIDErr` and `orgUUIDErr` are only used if one if statement and you don't have to personalize them because they have no impact further down the road. using and reusing `err` will do fine here.

systems/subscriber/registry/pkg/server/subscriber_test.go
```diff
@@ -0,0 +1,96 @@
...
+	msgbusClient.On("PublishRequest", mock.Anything, psub).Return(nil).Once()
+
+	s := server.NewSubscriberServer(subRepo, msgbusClient, nil, nil)
+	_, err := s.Update(context.TODO(), psub)
+
+	assert.NoError(t, err)
+	subRepo.AssertExpectations(t)
+}
+
+func TestLookupServer_GetOrg(t *testing.T) {
```
delvatt, line 72: Update the name

delvatt (changes requested):
There are still comments to address, including the nil pointer de-references. We will try to sort these out together tomorrow.

systems/subscriber/registry/pkg/client/network.go
```diff
@@ -0,0 +1,96 @@
...
+	if !resp.IsSuccess() {
+		logrus.Tracef("Failed to fetch network info. HTTP resp code %d and Error message is %s", resp.StatusCode(), errStatus.Message)
+		return fmt.Errorf(" Network Info failure %s", errStatus.Message)
+	}
+
+	err = json.Unmarshal(resp.Body(), &network)
+	if err != nil {
+		logrus.Tracef("Failed to desrialize network info. Error message is %s", err.Error())
+		return fmt.Errorf("network info deserailization failure:" + err.Error())
+	} else {
```
delvatt, line 66: You can remove the else block and indent back the logrus line to the same level of the if statements.

systems/subscriber/registry/pkg/db/subscriber_repo_test.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func TestSubscriber_Get(t *testing.T) {
+
+	t.Run("SubscriberFound", func(t *testing.T) {
+		var subID = uuid.NewV4()
+
+		// Arrange
+		db, mock, err := sqlmock.New()
+		if err != nil {
```
delvatt, line 116: you can use assert.NoError here

systems/subscriber/registry/pkg/server/subscriber.go
```diff
@@ -144,46 +139,59 @@ func (s *SubcriberServer) ListSubscribers(ctx context.Context, req *pb.ListSubsc
...
+		return nil, err
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
-
 	simRep, err := simManagerClient.ListSims(ctx, &simMangerPb.ListSimsRequest{})
 	if err != nil {
-		log.Fatalf("Failed to get Sims by subscriber. Error: %v", err)
+		logrus.Errorf("Failed to get Sims by subscriber. Error: %s", err.Error())
```
delvatt, line 148: handle the error

systems/subscriber/registry/pkg/server/subscriber_test.go
```diff
@@ -0,0 +1,52 @@
+package server
```
delvatt, line 1: We might need more tests.

systems/subscriber/registry/pkg/server/subscriber.go
```diff
@@ -144,46 +139,59 @@ func (s *SubcriberServer) ListSubscribers(ctx context.Context, req *pb.ListSubsc
...
 	}
+
 	allSims := simRep.Sims
 
+	// Store Sims by their SubscriberID
+	simMap := make(map[string][]*pb.Sim)
```
delvatt, line 154: Good!

systems/subscriber/registry/pkg/server/subscriber.go
```diff
@@ -144,46 +139,59 @@ func (s *SubcriberServer) ListSubscribers(ctx context.Context, req *pb.ListSubsc
...
+			FirstActivatedOn:   sim.FirstActivatedOn,
+			LastActivatedOn:    sim.LastActivatedOn,
+			ActivationsCount:   sim.ActivationsCount,
+			DeactivationsCount: sim.DeactivationsCount,
+			AllocatedAt:        sim.AllocatedAt,
+		})
+	}
+
 	var res []*pb.Subscriber
 	for _, sub := range subscribers {
```
delvatt, line 180: Seems good so far.

delvatt (approved):
Maybe add more test to the server, but seems good to me so far.  If I see anything later on, I'll open dedicated issue.

PR #402, opened 2023-02-07
Ukama agent node gateway

vthakur7f:


systems/ukama-agent/node-gateway/Int.Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM 003664043471.dkr.ecr.us-east-1.amazonaws.com/e2e-base-image:latest
+
+ENV PROJECT_NAME=api-gateway
```
delvatt, line 3: shouldn't the project name be `node-gateway` ?

delvatt (approved):

systems/ukama-agent/node-gateway/Int.Dockerfile
vthakur7f: fixed

PR #406, opened 2023-02-08
Subscriber system sim manager operator (telna) binding. 

### LINKED ISSUE #405: Implement the binding between telna agent and sim manager.
In order to send request to the telna agent, we need to implement a binding mechanism which consist of implementing a rest client of the telna system and wrap it into an adapter to the sim manager agent interface.

delvatt:
This pull request fixes #405


vthakur7f (approved):
Nice Work ..!!

PR #411, opened 2023-02-14
Ukama agent profile

vthakur7f:
Ukama-agent profile srevice added.


systems/ukama-agent/node-gateway/pkg/global.go
```diff
@@ -0,0 +1,5 @@
+package pkg
+
+const ServiceName = "node-gateway"
```
delvatt, line 3: `node_gateway`

systems/ukama-agent/node-gateway/pkg/global.go
```diff
@@ -0,0 +1,5 @@
...
+const SystemName = "ukama-gateway"
```
delvatt, line 4: `ukama_gateway`

systems/ukama-agent/profile/pb/profile.proto
```diff
@@ -0,0 +1,127 @@
+syntax = "proto3";
+
+option go_package = "pb/gen";
+
+package ukama.subscriber.profile.v1;
```
delvatt, line 5: if the agent is its own system, shouldn't the proto package namespace  be `package ukama.ukama_agent.profile.v1`;  instead of  `package ukama.subscriber.profile.v1;`

delvatt (approved):
LGTM. But the review was about both node-gateway and agent-profile.

systems/ukama-agent/node-gateway/pkg/global.go
vthakur7f: fixed

systems/ukama-agent/node-gateway/pkg/global.go
vthakur7f: fixed

systems/ukama-agent/profile/pb/profile.proto
vthakur7f: fixed

Issue #415, opened 2023-02-23
Integrate Prometheus Pushgateway to expose sim metrics


Brackleycassinga:
## Description:

We need to integrate Prometheus Pushgateway into our system to monitor the Sim metrics in the "subscribers" job. The metrics we want to monitor are the total number of subscribers, activation count, and deactivation count, which should be exposed as Gauge types. The activationCount and deactivationCount should be passed as constant labels to provide additional categorization and analysis of the data.

To push the metrics to the Pushgateway, we have implemented a helper function called PushMetrics. The function takes several parameters such as the metric job name, metric name, metric help text, metric type, metric labels, and metric value. Here is the function definition:

`func PushMetrics(metricJob, metricName string, metricHelp string, metricType prometheus.ValueType, metricLabels map[string]string, metricValue float64)
`
Currently, we are using the following code snippet to push metrics to the Pushgateway:

`utils.PushMetrics("subcribers", "number_of_subscribers", "The total number of subscribers", prometheus.GaugeValue, map[string]string{"activationsCount": strconv.FormatInt(activeCount, 10),"deactivationsCount": strconv.FormatInt(deactiveCount, 10)},  float64(simsCount))`

## KPIs Exposed:

- Number of subscribers: This metric measures the total number of subscribers currently subscribed to the service.

- Activation count: This metric measures the total number of subscriber activations that have occurred since the Pushgateway was last reset.

- Deactivation count: This metric measures the total number of subscriber deactivations that have occurred since the Pushgateway was last reset.

All of these metrics are exposed as Gauge types, which measure the value of a particular variable at a given point in time. In addition, the Activation Count and Deactivation Count metrics are passed as constant labels, allowing us to further categorize and analyze the data.

To retrieve these metrics from the database, we have implemented the simCount method which returns all the counts exposed by the Pushgateway.

PR #416, opened 2023-02-23
Integrate Prometheus Pushgateway for Sim Metrics

Brackleycassinga:
This pull request adds the integration of Prometheus Pushgateway into our sim manager service to monitor Sim metrics, including the total number of subscribers, activation count, and deactivation count. 

The metrics are exposed as Gauge types with constant labels, allowing us to further categorize and analyze the data. It also includes a new simCount method to retrieve current sim kpis data from the database.

issue #415


systems/subscriber/sim-manager/pkg/utils/prometheus_utils.go
```diff
@@ -0,0 +1,46 @@
...
+func PushMetrics(metricJob, metricName string, metricHelp string, metricType prometheus.ValueType, metricLabels map[string]string, metricValue float64) {
+	labelDimensions := make([]string, 0, len(metricLabels))
+
+	for key := range metricLabels {
+		labelDimensions = append(labelDimensions, key)
+	}
+
+	var metric prometheus.Collector
+
+	switch metricType {
```
vthakur7f, line 18: Here you would be creating metric collector for each push which is not right. May be try to maintain list of metric collectors which you have created. So before pushing check if metric collector exist. If it exisst get the metric collector and set the value otherwise create a new metric collector.

systems/subscriber/sim-manager/pkg/utils/prometheus_utils.go
```diff
@@ -0,0 +1,46 @@
...
+		metric = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: metricName,
+			Help: metricHelp,
+		}, labelDimensions)
+	default:
+		logrus.Errorf("Unsupported metric type: %v", metricType)
+		return
+	}
+
+	switch metricType {
```
vthakur7f, line 34: These function like creating and setting value could be reused from the exporter service file.

systems/subscriber/sim-manager/pkg/db/sim_repo.go
```diff
@@ -124,3 +125,21 @@ func (s *simRepo) Delete(simID uuid.UUID, nestedFunc func(uuid.UUID, *gorm.DB) e
...
+		return
+	}
+
+	result = s.Db.GetGormDb().Model(&Sim{}).Where("status =?", 2).Count(&activeCount)
+
+	if result.Error != nil {
+		err = result.Error
+		return
+	}
+	deactive = simsCount - activeCount
```
vthakur7f, line 143: Do we need a state for sim allocation which is for activation ? I am not sure.

vthakur7f (changes requested):

systems/subscriber/sim-manager/pkg/db/sim_repo.go
Brackleycassinga: Here, we retrieve the total number of SIMs and subtract the number of activated SIMs to obtain the count of deactivated SIMs. The fields `ActivationsCount` and `DeactivationsCount`, which are for a single SIM, cannot be used in this case.

systems/subscriber/sim-manager/pkg/utils/prometheus_utils.go
Brackleycassinga: Done

systems/subscriber/sim-manager/pkg/db/sim_repo.go
vthakur7f: I mant to say if you add more states in that case this will not be right ```deactive = simsCount - activeCount```

Brackleycassinga: Yes, you're correct. I will either separate the db calls or include a switch case to verify the state and return the appropriate value.

systems/subscriber/sim-manager/cmd/server/main.go
```diff
@@ -122,7 +122,7 @@ func runGrpcServer(gormDB sql.Db) {
 	grpcServer.ExtraUnaryInterceptors = []grpc.UnaryServerInterceptor{
 		fsInterceptor.UnaryServerInterceptor}
 
-	go msgBusListener(mbClient)
+	// go msgBusListener(mbClient)
```
vthakur7f, line 125: Does this work ? Are you able to publish messages without registering ?

systems/subscriber/sim-manager/pkg/db/sim_repo.go
```diff
@@ -124,3 +125,24 @@ func (s *simRepo) Delete(simID uuid.UUID, nestedFunc func(uuid.UUID, *gorm.DB) e
...
 	return err
 }
+func (s *simRepo) GetSimCounts() (simsCount, activeCount, deactiveCount, terminatedCount int64, err error) {
+	db := s.Db.GetGormDb()
+
+	if err := db.Model(&Sim{}).Count(&simsCount).Error; err != nil {
+		return 0, 0, 0, 0, err
+	}
+
+	if err := db.Model(&Sim{}).Where("status = ?", 1).Count(&activeCount).Error; err != nil {
```
vthakur7f, line 135: Can we use sim status constants rather than 1,2 or 3 ?

systems/subscriber/sim-manager/pkg/server/sim_manager.go
```diff
@@ -252,6 +262,22 @@ func (s *SimManagerServer) GetSimsBySubscriber(ctx context.Context, req *pb.GetS
 	return resp, nil
 }
 
+func (s *SimManagerServer) GetSimCounts(ctx context.Context) error {
+	terminatedCount, simsCount, activeCount, deactiveCount, err := s.simRepo.GetSimCounts()
+	if err != nil {
+		log.Errorf("failed to get Sims counts: %s", err.Error())
+		return err
+	}
+	customLabels := map[string]string{
```
vthakur7f, line 271: Kind of confused here why the Metric/KPI data is set as label here ? Active/Inactive/Terminated should be a metric on its own, not label.

vthakur7f (changes requested):
shaping up good but still needs some more changes.

systems/subscriber/sim-manager/pkg/db/sim_repo.go
Brackleycassinga: Yes, I can reuse the constant declared already on db model.

systems/subscriber/sim-manager/cmd/server/main.go
Brackleycassinga: I forgot to uncomment it after testing, which was my mistake. It has now been resolved.

systems/subscriber/sim-manager/pkg/server/sim_manager.go
Brackleycassinga: I have updated the Pushgateway util function,  now we are able to retrieve metrics for all KPIs instead of just labels.

systems/subscriber/sim-manager/pkg/utils/prometheus_utils.go
```diff
@@ -0,0 +1,46 @@
...
+	case prometheus.CounterValue:
+		metric.(*prometheus.CounterVec).With(prometheus.Labels(metricLabels)).Add(metricValue)
+	case prometheus.GaugeValue:
+		metric.(*prometheus.GaugeVec).With(prometheus.Labels(metricLabels)).Set(metricValue)
+	}
+
+	if err := push.New("http://localhost:9091/", metricJob).
```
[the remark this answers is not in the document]
Brackleycassinga: I added pushMetricHost to config.

systems/subscriber/sim-manager/pkg/utils/prometheus_utils.go
```diff
@@ -159,4 +159,25 @@ func PushMetrics(metricJob string, metrics []pkg.SimMetrics) {
     if err := pusher.Push(); err != nil {
```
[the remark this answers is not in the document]
Brackleycassinga: Addressed.

systems/subscriber/sim-manager/pkg/server/sim_manager.go
```diff
@@ -203,7 +205,10 @@ func (s *SimManagerServer) AllocateSim(ctx context.Context, req *pb.AllocateSimR
 	}
 
 	resp := &pb.AllocateSimResponse{Sim: dbSimToPbSim(sim)}
-
+	err = s.GetSimCounts(ctx)
```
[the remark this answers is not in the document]
Brackleycassinga: I have renamed the function from `GetSimCounts` to `GetSimMetrics` to more accurately reflect the data that the function returns.

systems/subscriber/sim-manager/pkg/server/sim_manager.go
```diff
@@ -252,6 +261,57 @@ func (s *SimManagerServer) GetSimsBySubscriber(ctx context.Context, req *pb.GetS
...
+		{
+			Name: "number_of_subscriber",
+			Type: "gauge",
+			Labels: map[string]string{
+				"type": "total",
+			},
+			Value: float64(simsCount),
+		},
+		{
+			Name: "number_of_subscriber",
```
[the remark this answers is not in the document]
Brackleycassinga: This was fixed on old commit

systems/subscriber/sim-manager/pkg/server/sim_manager.go
```diff
@@ -229,6 +234,10 @@ func (s *SimManagerServer) GetSim(ctx context.Context, req *pb.GetSimRequest) (*
 }
 
 func (s *SimManagerServer) ListSims(ctx context.Context, req *pb.ListSimsRequest) (*pb.ListSimsResponse, error) {
+	err := s.GetSimCounts(ctx)
+	if err != nil {
+		log.Errorf("Error while pushing sim metric to pushgatway %s", err.Error())
```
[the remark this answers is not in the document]
Brackleycassinga: Addressed.

systems/subscriber/sim-manager/mocks/RunningOrgClient.go
vthakur7f, line 1: Mocks require clean up.

systems/common/pushgatewayMetrics/pushgatewayMetrics.go
vthakur7f, line 1: I think this file should be part of metrics package. I hope some of the functions used in this file will be reused in exporter too.

systems/common/pushgatewayMetrics/pushgatewayMetrics.go
```diff
@@ -0,0 +1,192 @@
...
+			pusher.Collector(m.collector)
+		}
+	}
+	if err := pusher.Push(); err != nil {
+		log.Errorf("Could not push metrics to Pushgateway: %s", err.Error())
+	}
+}
+
+func CollectAndPushSimMetrics(pusMetricHost string, configMetrics []MetricConfig, selectedMetric string, Value float64, Labels map[string]string, systemName string) error {
+	var selectedMetrics []MetricConfig
```
vthakur7f, line 169: Does this needs to be an array fi we are only getting one metric?

systems/common/pushgatewayMetrics/pushgatewayMetrics.go
```diff
@@ -0,0 +1,192 @@
...
+		for _, m := range metrics {
+			pusher.Collector(m.collector)
+		}
+	}
+	if err := pusher.Push(); err != nil {
+		log.Errorf("Could not push metrics to Pushgateway: %s", err.Error())
+	}
+}
+
+func CollectAndPushSimMetrics(pusMetricHost string, configMetrics []MetricConfig, selectedMetric string, Value float64, Labels map[string]string, systemName string) error {
```
vthakur7f, line 168: pusMetricHost -> pushGatewayHost

vthakur7f (commented):
Couple of small changes required before it can be rebase to main line metric system branch.

systems/common/pushgatewayMetrics/pushgatewayMetrics.go
Brackleycassinga: Yes, that's correct. Since we've set up the "before send" function in `config`, it's necessary as the config as already all the labels. If we change it from an array, we'll have to manually set each metric for every request.

systems/common/pushgatewayMetrics/pushgatewayMetrics.go
Brackleycassinga: Addressed

systems/common/pushgatewayMetrics/pushgatewayMetrics.go
vthakur7f: var selectedMetrics []MetricConfig  this is anyway a local variable and will be set every time you call CollectAndPushSimMetrics function. Since you are calling this with one metric name argument " selectedMetric string" which means at a time only one metric can be pushed. So I don't see a need of array. Let me know If I am missing anuything.

Brackleycassinga: You right.

Brackleycassinga: You're right .It need to be a single metric .

vthakur7f (approved):

PR #423, opened 2023-02-28
Registry system users updates. fixes #422

### LINKED ISSUE #422: Update Users to match new set of standards and conventions
- Use UUID v4 instead of uint64 for IDs
- Use our common/uuid  instead of any other specific library
- Alias all logs to logrus

delvatt:
This pull request fixes #422 and closes #422


systems/registry/users/pb/users.proto
```diff
@@ -26,15 +26,15 @@ message AddResponse{
...
+    string userUuid = 1 [(validator.field) = {uuid_ver: 4, string_not_empty: true}, json_name = "user_uuid"];
 }
 
 message GetResponse {
     User user = 1;
 }
 
 message UpdateRequest {
-    string userUuid = 1 [(validator.field) = {uuid_ver: 0, string_not_empty: true}, json_name = "user_uuid"];
+    string userUuid = 1 [(validator.field) = {uuid_ver: 4, string_not_empty: true}, json_name = "user_uuid"];
```
Brackleycassinga, line 37: Instead of using `useruuid`, which is already known to be a UUID, let's use 'userID' to make the variable name more concise.

systems/registry/users/pb/users.proto
```diff
@@ -67,7 +67,7 @@ message User {
     string name = 1 [(validator.field) = { string_not_empty: true,  length_gt: 1 }];
     string email = 2  [(validator.field) = { human_error:"must be an email format" regex: "^$|^[a-z0-9._%+\\-]+@[a-z0-9.\\-]+\\.[a-z]{2,4}$"}];
     string phone = 3  [(validator.field) = { human_error:"must be a phone number format", regex: "^$|^(?:(?:\\(?(?:00|\\+)([1-4]\\d\\d|[1-9]\\d?)\\)?)?[\\-\\.\\ \\\\\\/]?)?((?:\\(?\\d{1,}\\)?[\\-\\.\\ \\\\\\/]?){0,})(?:[\\-\\.\\ \\\\\\/]?(?:#|ext\\.?|extension|x)[\\-\\.\\ \\\\\\/]?(\\d+))?$"}];
-    string uuid = 4  [(validator.field) = {uuid_ver: 0}];
+    string uuid = 4  [(validator.field) = {uuid_ver: 4}];
```
Brackleycassinga, line 70: Can we have conventional naming for ID field .I think it is better to rename this to `userID` or `ID`.

Brackleycassinga (changes requested):

systems/registry/users/pb/users.proto
delvatt: Thanks. But there's already a User.ID (because I was using gorm.Model)  which will translate into UserID in DB foreign keys. That's why I've kept the UserUuid.

systems/registry/users/pb/users.proto
delvatt: We have this here because we are already using an ID field.

Brackleycassinga (approved):

Issue #418, opened 2023-02-27
Enhance the registry system with message client integration to publish events on certain actions


Brackleycassinga:
## Description:
The current pull request registry-system-msgClient-integration proposes an integration of message client to publish events when certain actions are performed on the registry system. This will enhance the system's functionality by enabling it to trigger events in response to actions such as creating a network, adding an organization, adding a member to an organization, deleting an organization, or deleting a user. The integration covers all the four services in the system, including Org, Users, Network, and Nodes.

This integration will enable users to monitor the system's activities in real-time and respond accordingly. For instance, they can receive notifications on the creation of a new network or the addition of a new user to an organization. Such notifications will help users to stay up to date with the system's activities and respond promptly to any issues that may arise.

list of exposed events:

## Node events:

addnode
attach node
detach node

## Users events:
add
delete
deactivate

## Org events:
add
delete
add member
remove member

## Network events:
add
delete

PR #425, opened 2023-03-01
Org msgclient integration

Brackleycassinga:
#418


systems/registry/org/org_reg.patch
```diff
@@ -0,0 +1,2424 @@
+diff --git a/systems/registry/org/cmd/server/main.go b/systems/registry/org/cmd/server/main.go
```
delvatt, line 1: this file should be removed  and checked out of git repo  (git rm)

delvatt (approved):
LGTM! But do not forget to remove (un-stage and delete the patch file with git rm)

systems/registry/org/org_reg.patch
Brackleycassinga: Addressed

PR #439, opened 2023-03-07
Registry, Data-plan and subscriber systems naming convention fixes

salman-codewar:
## Purpose changes

- Naming convention fixes in Subscriber, Registry and DataPlan systems
- Updated Date inputs to RF1123 standard for DataPlan and Subscriber system
- Added example values in  API structs for easy of testing in Data plan, Registry and Subscriber API GW's
- Minor fixes


systems/data-plan/base-rate/pkg/validations/validations.go
```diff
@@ -17,10 +18,23 @@ func IsValidUploadReqArgs(fileUrl, effectiveAt, simType string) bool {
...
-func IsFutureDate(date string) bool {
-	t, err := time.Parse(time.RFC3339, date)
-	if err != nil {
-		return false
-	}
-	return time.Now().Before(t)
+
+
+func IsFutureDate(date string) error {
+    t, err := time.Parse(time.RFC1123, date)
```
vthakur7f, line 24: Idea is to have to have time methods in common package and use it from there. But I think @delvatt  is working on that. May be once he is done will update this too.

vthakur7f (approved):

PR #440, opened 2023-03-08
Metrics system exporter

vthakur7f:
Exporter service to present metrics for scraping to prometheus.
Exporter utilizes events from the systems and present the metrics for prometheus to scrape based on the config supplied.


vthakur7f:
sim.proto file will be updated once @delvatt Pr gets merged.

delvatt (approved):
LGTM!

PR #441, opened 2023-03-10
Updated service name to single word.

vthakur7f:
ServiceNames are causing lot off issues in deployment and testing.
1) Service names are getting used in rotuing keys.
2) Env varibles to read service port IP.
3) Deployments for creating name spaces.

We need to make sure all the constrains imposed by these are met. 
So simplest way for now is to use a single word name as it would satisfy rotuing key constraint and still meet our needs.


salman-codewar (approved):

salman-codewar:
LGTM

delvatt (approved):
LGTM!

PR #446, opened 2023-03-14
Subscriber system: sim manager operator sim auto activate

### LINKED ISSUE #444: Implement auto activate for operator (telna) sims.
Some operator's sims are automatically set as active on first use. In order to support this behavior we also need to implement some kind of auto activation (event based) for these type of sims.

delvatt:
This PR fixes #444  and closes #444


vthakur7f (approved):
Thanks..!! Great work.

PR #449, opened 2023-03-14
Bff system node api updates ->Api gateway node endpoint added

Brackleycassinga:
Issue #450


systems/docker-compose.yml
```diff
@@ -1,15 +1,13 @@
 version: "3.9"
```
salman-codewar, line 1: I think we should revert these docker-compose changes

systems/registry/api-gateway/pkg/rest/api.go
```diff
@@ -5,10 +5,25 @@ package rest
 type GetOrgsRequest struct {
 	UserUUID string `form:"user_uuid" json:"user_uuid" query:"user_uuid" binding:"required" validate:"required"`
 }
+type GetNodeRequest struct {
```
salman-codewar, line 8: We can also add example values on these structs, Like
`example:"brackley-node"`, that way it'll be pre-populated in Postman collection and swagger

salman-codewar (changes requested):

salman-codewar:
Good Job man, Added some comments. 👍 
Also PR have some conflicts, i think you need to rebase it with `bff-system`

salman-codewar (approved):

PR #452, opened 2023-03-16
Main msgclient updates

vthakur7f:
MsgClient:
- Moved msgClient to messaging system.
- Updated all services which were using message client.
- Regenrated mocks and updated all dependecies which were using init/msgclient.


salman-codewar:
Thanks man for this improvement and updates throughout out the systems 👍

salman-codewar (approved):

PR #457, opened 2023-03-18
Frontend system > Data-plan - Packages endpoint integrated

salman-codewar:


systems/bff/src/modules/packages/service.ts
```diff
@@ -0,0 +1,84 @@
...
+import {
+    AddPackageInputDto,
+    PackageDto,
+    PackagesResDto,
+    UpdatePackageInputDto,
+} from "./types";
+@Service()
+export class PackageService implements IPackageService {
+    getPackage = async (
+        packageId: string,
```
Brackleycassinga, line 18: The PackageId parameter is being passed but not utilized.
`pkgId` need to passed as a query params in url .

systems/bff/src/modules/packages/service.ts
```diff
@@ -0,0 +1,84 @@
...
+            type: API_METHOD_TYPE.GET,
+            path: `${SERVER.DATA_PLAN_PACKAGES_API_URL}/org/${cookie.orgId}`,
+            headers: cookie.header,
+        });
+        if (checkError(res)) throw new Error(res.message);
+        if (!res) throw new HTTP404Error(Messages.NODES_NOT_FOUND);
+        return PackageMapper.dtoToPackagesDto(res);
+    };
+    addPackage = async (
+        req: AddPackageInputDto,
```
Brackleycassinga, line 41: There is a compilation error input type for `duration` of `AddPackageInputDto`

Brackleycassinga (changes requested):
Minor changes are needed, but otherwise, it's good to go.

systems/bff/src/modules/packages/service.ts
salman-codewar: Addressed

Brackleycassinga (approved):
Looks good to me.

PR #474, opened 2023-03-28
Billing system Collector

### LINKED ISSUE #473: Implement exporter
Implement the core exporter service within the billing system that will collect billable events from all other systems.

delvatt:
This pull request fixes #473 and closes #473


vthakur7f (commented):
code looks good. Unit  test cases are missing for server and client. I think we could skip client but server can be added.

delvatt:
> code looks good. Unit test cases are missing for server and client. I think we could skip client but server can be added.

Hey Vishal, thanks for the feedback.  Yes, per our standards, units tests for the server would have been added before merging.  The main idea was to get the exporter design and  code out early so that you could have a look.  I'll be fixing the conflicts and add the tests shortly.

vthakur7f (commented):
Unit test still missing

delvatt:
@vthakur7f  Thanks for the reminder.  unit tests added.

systems/billing/collector/internal/server/event.go
```diff
@@ -0,0 +1,280 @@
...
+			return nil, err
+		}
+
+		err = handleRegistrySubscriberUpdateEvent(e.RoutingKey, msg, b)
+		if err != nil {
+			return nil, err
+		}
+
+	// Delete customer
+	case "event.cloud.registry.subscriber.delete":
```
vthakur7f, line 79: It's kind of confusing I think we need to add system name to events

systems/billing/collector/internal/server/event.go
```diff
@@ -0,0 +1,280 @@
...
+	defer cancel()
+
+	event := client.Event{
+		//TODO: To be replaced by msgClient msgId
+		TransactionId: fmt.Sprintf("%s%d", simUsage.Id, time.Now().Unix()),
+
+		CustomerId:     simUsage.SubscriberId,
+		SubscriptionId: simUsage.SimId,
+		Code:           "data_usage",
+		SentAt:         time.Now(),
```
vthakur7f, line 122: What is this time used for ? Hopefully not in the user invoice.

systems/billing/collector/internal/server/event.go
```diff
@@ -0,0 +1,280 @@
...
+			"sim_id":     simUsage.SimId,
+		},
+	}
+
+	log.Infof("Sending operator data usage event %v to billing server", event)
+
+	return b.client.AddUsageEvent(ctx, event)
+}
+
+func handleRegistrySubscriberCreateEvent(key string, subscriber *subpb.Subscriber,
```
vthakur7f, line 135: I think we are still missing a piece where we confirm we have payment info for the subscriber. That should be done before activating/allocating sim.

vthakur7f (approved):
Check the conflict. Otherwise looks ok. I think there are few things which would need a closer look in future.

systems/billing/collector/internal/server/event.go
delvatt: > It's kind of confusing I think we need to add system name to events

I think the confusion comes from the fact that the **subscriber-registry** service (inside the **subscriber system)** was renamed `registry`, when we already have a system that is also called `registry`.

systems/billing/collector/internal/server/event.go
delvatt: > What is this time used for ? Hopefully not in the user invoice.

This is when the usage event is sent, and I am not sure but I think it might be used for invoice because invoice is calculated  for events that were sent within the billable period. Will  definitely have to check.

systems/billing/collector/internal/server/event.go
delvatt: > I think we are still missing a piece where we confirm we have payment info for the subscriber. That should be done before activating/allocating sim.

Yeah. That is still a missing piece in the subscriber system.

systems/billing/collector/internal/server/event.go
delvatt: > > What is this time used for ? Hopefully not in the user invoice.
> 
> This is when the usage event is sent, and I am not sure but I think it might be used for invoice because invoice is calculated for events that were sent within the billable period. Will definitely have to check.

Checking or not, either way  now I think this should be mapped to the ending time of the data session.  We'll update in the next iteration.

Issue #477, opened 2023-03-28
Implement org rate service


vthakur7f:
Implement Org rate service. This service would provide the actual rates applicable for subscriber per package.
This would provide interface to add org default markup and markup per package.   

- [x] Org Rate service
- [x] API gateway updates
- [x] Base rate updates
- [x] CI/CD Work 
- [x] Docker Compose for test

PR #493, opened 2023-04-10
Data plan system rate

vthakur7f:
Fixes issue #477


systems/data-plan/api-gateway/pkg/client/dataPlan.go
```diff
@@ -5,23 +5,20 @@ import (
 	"time"
```
salman-codewar, line 5: I think we can rename this file to `package.go`

salman-codewar (approved):

salman-codewar:
Just added one small comment, Rest looks great 👍 💯

delvatt (approved):
LGTM!

systems/data-plan/api-gateway/pkg/client/dataPlan.go
vthakur7f: yeah I think I missed it somehow.

Issue #487, opened 2023-04-04
Integrate auth middleware in systems


salman-codewar:
To authenticate incoming request to systems we need to integrate auth middleware within systems. Auth middleware will get the session data from request headers and validate it from identity manager.

- [x] Implement **auth provider** in `systems/common/provider`. Provider will provide `authClient` instance and `authenticateUser` function.

Once the auth method added to common we need to integrate in systems:

- [x] Integrate auth middleware in `registry` 
- [x] Integrate auth middleware in `data-plan` 
- [x] Integrate auth middleware in `subscriber`
- [x] Integrate auth middleware in `registry`  

On successful `GetUserBySession` response proceed to request, otherwise return `unauthorised` response. 

- [x] Add github workflow for auth system
- [x] Auth system deployment

Issue #480, opened 2023-03-30
Implement Auth system


Brackleycassinga:
## Description:
As our application grows, we need to have a more fine-grained control over the access and actions that users can perform. To achieve this, we need to implement a role-based authorization system using Ory Kratos .

- The authorization system should allow us to define different roles and assign them to users. Each role should have a set of permissions that determine the actions that users with that role can perform.

- We should also be able to define tasks, which are specific actions that users can perform. Each task should be associated with one or more roles, and only users with those roles should be able to perform the task.

- To achieve this, we need to create a new microservice that integrates with Ory Kratos and provides an API for managing roles . The microservice should use Ory Kratos for authentication and authorization, and it should store role and task information in a database.

Tasks:

- [x] Implement Auth API-GW
- [ ] Ory Keto deployment and confuguration

PR #497, opened 2023-04-11
Auth system

salman-codewar:
## Purpose changes

Covers issue: #487 #480 

**Systems - Common**

Add Auth Provider in common, which will:
- [x]  Provide auth client instance
- [x] Provide authenticateUser method

**Systems - Auth/ApiGateway**

- [x] Auth system API GW added
- [x] Added Auth app redirect on swagger UI to et cookies in browser

**Systems - Auth Middleware integration**

- [x] Auth Middleware integration in Data Plan system
- [x] Auth middleware integration in Subscriber system
- [x] Auth middleware integration in Registry
- [x] Minor refactoring
- [x] Deployment


systems/data-plan/api-gateway/pkg/rest/router.go
```diff
@@ -19,6 +20,8 @@ import (
 	"github.com/wI2L/fizz/openapi"
 )
 
+var REDIRECT_URI = "https://data-plan.dev.ukama.com/swagger/#/"
```
vthakur7f, line 23: This doesn't look good. what happnes when somebody else deploy this for their org ?

vthakur7f (commented):
Only thing I can point here is hardcoding of REDIRECT_URI add org name won't wokr for us. Lets get rid of these. After that we will be good to go.

systems/data-plan/api-gateway/pkg/rest/router.go
salman-codewar: Thanks for the feedback @vthakur7f , Addressed

vthakur7f (approved):
Nice Work..!!!

PR #530, opened 2023-06-09
Mailing system

Brackleycassinga:
Added a REST interface for seamless communication with our AWS mailing service. It acts as a bridge between Ukama systems and AWS, enabling efficient and reliable email sending.


systems/common/config/config.go
```diff
@@ -230,6 +238,58 @@ func LoadServiceHostConfig(name string) *Service {
 
 	return s
 }
+func LoadMailerHostConfig(name string) *Mailer {
```
vthakur7f, line 241: Do we have multiple services which would be using this LoadMailerHostConfig or it's just a mailer config ?

systems/notification/mailer/pkg/server/mailer.go
```diff
@@ -0,0 +1,129 @@
...
+	Values  map[string]interface{}
+}
+
+type MaillingServer struct {
+	maillingRepoRepo db.MaillingRepo
+	pb.UnimplementedMaillingServiceServer
+	mailer *config.Mailer
+}
+
+func NewMaillingServer(maillingRepoRepo db.MaillingRepo, mail *config.Mailer) *MaillingServer {
```
vthakur7f, line 35: *Mailing Need to fix in at most of method name, proto and struct .

systems/notification/mailer/pkg/db/model.go
```diff
@@ -0,0 +1,19 @@
+package db
+
+import (
+	"time"
+
+	uuid "github.com/ukama/ukama/systems/common/uuid"
+)
+
+type Mailing struct {
+	MailId    uuid.UUID  `gorm:"primaryKey;type:uuid"`
```
vthakur7f, line 10: Should this mailId be sent in email also for reciever so that he can refer to Id in further communication if required ?

systems/notification/mailer/pkg/db/mailer_repo.go
```diff
@@ -0,0 +1,24 @@
...
+	Db sql.Db
+}
+
+func NewMaillingRepo(db sql.Db) MaillingRepo {
+	return &maillingRepo{
+		Db: db,
+	}
+}
+
+func (s *maillingRepo) SendEmail(mail *Mailing) error {
```
vthakur7f, line 21: Should we have way to pull communication based on certain criteria like email or userid ? It's totally fine to skip the implemenation now and have it in future. Just need to make sure we have that in issue list.

systems/notification/mailer/pb/mailer.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+option go_package = "pb/gen";
```
vthakur7f, line 2: Updated a go package "option go_package = "github.com/ukama/ukama/systems/notification/mailer/pb/gen";

vthakur7f (commented):
Great work..!! Few simple fixes and CI required otherwise looks good.

systems/notification/mailer/pkg/server/mailer.go
Brackleycassinga: I have updated all the names to "mailer" throughout the code.

systems/notification/mailer/pkg/db/model.go
Brackleycassinga: It makes sense to include it in the email body's default values so that users can easily add it to their email templates. We can include it in the email template this way:
`<p>Email ID: {{ .Values.EmailID }}</p>
`

systems/notification/mailer/pkg/db/mailer_repo.go
Brackleycassinga: Make sense .I will add a method to pull email by ID.

systems/notification/mailer/pb/mailer.proto
Brackleycassinga: Done.

vthakur7f (approved):
proto file go package is not updated. Please update it before merge.

PR #531, opened 2023-06-12
Messaging system nns

vthakur7f:
Adding
- [X] Node name server
- [X] Api Gateway


delvatt (approved):
Great PR, as always!  Thanks.
Just a small reminder as we agreed to alias `logrus`  imports to `log`.

vthakur7f:
> Reference
Updates. Thanks.

PR #538, opened 2023-06-29
Registry system node

vthakur7f:
Added node status for maintaining connectivity and node state.
Node Connectivity state can be any of follwing:
- Online
- Offline

Node State is represented by one of following values:
- Onboard
- Configured
- Maintenance
- Faulty
- Active


systems/registry/node/pkg/db/node_repo_test.go
```diff
@@ -163,71 +167,6 @@ func TestNodeRepo_Get(t *testing.T) {
 	})
 }
 
-func TestNodeRepo_GetForOrg(t *testing.T) {
```
delvatt, line 166: Is there any reasons why we are removing the test for NodeRepo_GetForOrg ?

systems/registry/node/pkg/db/nodestatus_repo.go
```diff
@@ -0,0 +1,92 @@
...
+
+func NewNodeStatusRepo(db sql.Db) NodeStatusRepo {
+	return &nodeStatusRepo{
+		Db: db,
+	}
+}
+
+func (n *nodeStatusRepo) Update(ns *NodeStatus) error {
+	err := n.Db.GetGormDb().Transaction(func(tx *gorm.DB) error {
+		t := tx.Where("node_id= ?", ns.NodeId).Delete(&NodeStatus{})
```
delvatt, line 30: Make sure you send lowercased nodeIds against the DB with ns.NodeId.StringLowercase().

systems/registry/node/pkg/db/nodestatus_repo.go
```diff
@@ -0,0 +1,92 @@
...
+	result := n.Db.GetGormDb().Find(&ns)
+
+	if result.Error != nil {
+		return nil, result.Error
+	}
+
+	return ns, nil
+}
+
+func (n *nodeStatusRepo) GetNodeCount() (activeNodeCount, inactiveNodeCount int64, err error) {
```
delvatt, line 80: shouldn't those be  onlineNodeCount and offlineNodeCount instead of activeNodeCount and inactiveNodeCount?

systems/registry/node/pkg/server/node.go
```diff
@@ -68,46 +70,20 @@ func (n *NodeServer) AddNode(ctx context.Context, req *pb.AddNodeRequest) (*pb.A
...
 
 	node := &db.Node{
 		Id:    req.NodeId,
-		OrgId: orgId,
-		State: nodeState,
-		Type:  nID.GetNodeType(),
-		Name:  req.Name,
+		OrgId: n.org,
+		Status: db.NodeStatus{
+			NodeId: req.NodeId,
```
delvatt, line 81: Make sure you send lowercased nodeIds against the DB.

systems/registry/node/pkg/db/model.go
```diff
@@ -10,28 +10,40 @@ import (
...
+	OrgId     uuid.UUID  `gorm:"type:uuid;not null"`
+	Attached  []*Node    `gorm:"many2many:attached_nodes"`
 	CreatedAt time.Time
 	UpdatedAt time.Time
 	DeletedAt gorm.DeletedAt `gorm:"index"`
 }
 
+type NodeStatus struct {
+	gorm.Model
+	NodeId string
```
delvatt, line 26: We should make sure NodeId is unique (and maybe indexed).

systems/registry/node/pkg/server/node.go
```diff
@@ -198,51 +174,45 @@ func (n *NodeServer) GetNodes(ctx context.Context, req *pb.GetNodesRequest) (*pb
...
-		NodeId: nodeUpdates.Id,
-		State:  nodeUpdates.State.String(),
+		return nil, grpc.SqlErrorToGrpc(err, "node")
 	}
 
-	err = n.msgbus.PublishRequest(route, evt)
+	und, err := n.nodeRepo.Get(nodeId)
 	if err != nil {
-		log.Errorf("Failed to publish message %+v with key %+v. Errors %s", evt, route, err.Error())
+		log.Error("error updating the node state, ", err.Error())
```
delvatt, line 213: I think  if we were able to successfully proceed to NodeStateUpdate(), but are failing to get back the corresponding node with NodeRepo.Get() after that, then this is even a more severe error than failing to update the node's state.  This probably means that the Node DB and the NodeState DB are in an inconsistent state in regard to each other.

systems/registry/node/pkg/server/node_test.go
```diff
@@ -21,28 +21,32 @@ import (
 )
```
delvatt, line 21: This file "systems/registry/node/pkg/server/node_test.go " reports few linting errors by GitHub Actions. If these linting errors are not reported in your IDE, make sure you check them on GitHub.

delvatt (approved):
Great job as always!!!
Some minor comments there and there. Also. GitHub Actions reports some lint errors from some files (including from those: systems/notification/mailer/test/integration/mailer_test.go, systems/notification/mailer/cmd/server/main.go). I am not sure why these are showing up in you PR, but if you can have a look, that will be fine.

systems/registry/node/pkg/db/node_repo_test.go
vthakur7f: Yeah we had to clean this as part of registry tear down. Have to figure out better way to do the verfication.

systems/registry/node/pkg/server/node.go
vthakur7f: Failed to get could be due to number of factors whicj are very unlikely to happen like db unreachable (nw failure) or that something that kind of stuff. I dont think adding err clause here was meant for like,issing node id or sometink like that.

systems/registry/node/pkg/server/node_test.go
vthakur7f: fixed

systems/registry/node/pkg/db/node_repo_test.go
delvatt: Agree!

PR #542, opened 2023-07-03
Subscriber system hotfixes

### LINKED ISSUE #541: Subscriber/Sim Manager:  Fix unbound sim between manager and agent.
After allocation, sim represented in sim Manager and sim represented in remote agent are not necessarily bound. This leads sometime to small bugs when trying to perform basic operations on sims from sim manager such as activate/deactivate sim. 

The solution is to bind the sim allocated from sim manager to the remote agent sim when this operation is supported by the remote agent, and perform a NO-OP operation for agents that do not need this.  All this is managed by the agents adapters, providers and factories interfaces, transparently, without any knowledge from the sim manager.

delvatt:
This PR fixes #541


Brackleycassinga (approved):
LGTM 👍🏾

PR #559, opened 2023-08-03
Registry system user restore proto

delvatt:
This PR  adds the following changes: 

-  Remove not empy requirement on userId  proto field
-  Remove old state param in AddNode interface.


Brackleycassinga (approved):
LGTM

PR #595, opened 2023-10-09
API System: Implement public APIs for non UI agents. Fixes #594

### LINKED ISSUE #594: Implement public APIs for non UI agents
the Ukama platform can be accessed by human operators via a UI console, but  should also be accessible for non human agents such as robots or other computer programs. In order for this to happen, the platform needs to expose a public set of APIs that can be instrumented by these non human agents. 

The initial set of APIs will allow to perform the following operations: 

- Create a new network
- Add a new package 
- Configure a  new sim for a given subscriber

delvatt:
This PR fixes #594 .  It adds the following set of features: 

- Create a new network
- Add a new package
- Configure a new sim for a subscriber


systems/api/api-gateway/pkg/client/client.go
```diff
@@ -0,0 +1,317 @@
+package client
```
salman-codewar, line 1: We have some similar kind of code repetition in clients, Where we Marshal req, send req and unmarshall response.
We had the similar structure in Testing system, I added a function to abstract that part. Have a look
https://github.com/ukama/ukama/blob/ed21058a74e0b7d8ab03cf9eedd0d017133336c8/testing/integration/pkg/utils/rest.go#L211-L300

And that's how we can call it

https://github.com/ukama/ukama/blob/ed21058a74e0b7d8ab03cf9eedd0d017133336c8/testing/integration/pkg/nucleus/nucleus.go#L53-L62

systems/api/api-gateway/pkg/config.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/ukama/ukama/systems/common/config"
+	"github.com/ukama/ukama/systems/common/rest"
+)
+
+type Config struct {
+	config.BaseConfig `mapstructure:",squash"`
+	DB                *config.Database `default:"{}"`
+	Server            rest.HttpConfig
+	// Services          GrpcEndpoints  `mapstructure:"services"`
+	// Service          *uconf.Service
```
salman-codewar, line 16: Some commented code in this file, Let's remove it if we don't need otherwise we can add TODO for ref.

salman-codewar (changes requested):
Awesome stuff man, mentioned couple of minor things. Good job 👍 💯

systems/api/api-gateway/pkg/config.go
delvatt: Addressed.

systems/api/api-gateway/pkg/client/client.go
delvatt: As discussed, we'll investigate the better way to perform this refactoring in another issue, so that it can benefit other systems.

salman-codewar (approved):

systems/api/api-gateway/pkg/client/client.go
```diff
@@ -0,0 +1,322 @@
...
+	GetNode(string) (*NodeInfo, error)
+	RegisterNode(string, string, string, string) (*NodeInfo, error)
+	AttachNode(string, string, string) error
+	DetachNode(string) error
+	AddNodeToSite(string, string, string) error
+	RemoveNodeFromSite(string) error
+	DeleteNode(string) error
+}
+
+type clients struct {
```
vthakur7f, line 39: I think right way would be have a interface per client Like interfca efor network, package, sim etc.
Rahther than haveing one cleint interface. May it could be useful at somepoint of time when we try to move all these interfaces as common package.

systems/api/api-gateway/pkg/client/package.go
vthakur7f, line 1: This looks almost duplicate of what we doing in testing framework. Hope we are reusing stuff.

systems/api/api-gateway/pkg/client/rest.go
vthakur7f, line 1: same this is also like tesing framework. Should we import these packages and use ?

systems/data-plan/package/pkg/server/package.go
```diff
@@ -186,6 +189,10 @@ func (p *PackageServer) Add(ctx context.Context, req *pb.AddPackageRequest) (*pb
 		PackageDetails: db.PackageDetails{
 			Apn: req.Apn,
 		},
+		Overdraft:     req.Overdraft,
+		TrafficPolicy: req.TrafficPolicy,
+		Networks:      req.Networks,
+		SyncStatus:    types.SyncStatusPending,
```
vthakur7f, line 195: What is this for ?

vthakur7f (commented):
Looks good to me. But looks like lot of code can be reused between testing and this api system.
Not sure whats the plan. @salman-codewar Can you have a look at it too.

delvatt:
> Looks good to me. But looks like lot of code can be reused between testing and this api system. Not sure whats the plan. @salman-codewar Can you have a look at it too.

That's what we were discussing in today's standup and we decided to have a more deeper look at that in a separate issue. There are some slight differences in needs between the two, especially especially how we handle responses for the APIs that we do not do in testing, that's why I did not abstracted the code in common yet.

salman-codewar:
> Looks good to me. But looks like lot of code can be reused between testing and this api system. Not sure whats the plan. @salman-codewar Can you have a look at it too.

Yea we need to plan this improvement. We can't reuse code between testing workflows and API because testing workflow is outside systems.
Here we have can improve 2 kind of reusability

1. Between testing workflows and Ukama API
That can't be done until we decided to make testing part of systems. On this @delvatt mentioned very valid point that we should not use same code for testing and api's implementation. So keep testing out of systems make more sense.

2. Between different API clients
Because API is communicating with all systems and communication can be of either protobuf of rest, So we should plan a way to common out these communication like (marshalling, un-marshalling and error handling). Since API is connecting will all systems we can expect this system to expand much faster, and if we didn't tackle reusability between API clients it'll bring lot of duplicate code very soon.

My opinion was to start with error handling, we can have custom error handing for API which will be just a wrapper and have custom error code which we can use between multiple clients rather adding same condition every time. These custom error code are separate from http status code.
Anyways we can discuss it more on call.

So my suggestion is we should not hold this PR because of this improvement, we can merge this PR and prioritise improvement task before next code addition in the system.

systems/api/api-gateway/pkg/client/client.go
delvatt: Addressed

systems/api/api-gateway/pkg/client/package.go
delvatt: we'll extract in a subsequent issue.

systems/api/api-gateway/pkg/client/rest.go
delvatt: we'll extract in a subsequent issue.

systems/data-plan/package/pkg/server/package.go
delvatt: We use that to update the resource status because the second part of the request is asynchronous (event based). We that in order to provide the right  resource status, whenever the caller request it.

PR #597, opened 2023-10-12
Node config

vthakur7f:
Initial Version of node config.
- [X] any json file under the org/network/site/node/ is considered as a config
- [X] Any changes to the config in git is collected by a REST event.
- [X] Events are collected and processed by configurator to check if config needs to be really pushed to nodes. 
- [X] Pushes the config to node using node feeder and collect the status back as an event.


systems/node/api-gateway/pkg/config.go
```diff
@@ -41,7 +42,7 @@ func NewConfig() *Config {
 			Controller: "controller:9090",
```
Brackleycassinga, line 42: Missing configurator svc

Brackleycassinga (approved):
LGTM 👍🏾

PR #605, opened 2023-10-26
registry/api-gateway: move state param from path to json payload. Fixes #600

### LINKED ISSUE #600: Registry Node > Update node state req is missing "state" in path param
**Describe the bug**
In `UpdateNodeState` req struct we are defined `state` as path param but its not defined in route

**Code Ref**

Struct
https://github.com/ukama/ukama/blob/ed21058a74e0b7d8ab03cf9eedd0d017133336c8/systems/registry/api-gateway/pkg/rest/api.go#L61-L64

Route

https://github.com/ukama/ukama/blob/ed21058a74e0b7d8ab03cf9eedd0d017133336c8/systems/registry/api-gateway/pkg/rest/router.go#L199-L199

delvatt:
This PR fixes #600  with the following change: 

- Move state param from path to json payload for NodeStateUpdate

### DIFF
--- a/systems/registry/api-gateway/pkg/rest/api.go
+++ b/systems/registry/api-gateway/pkg/rest/api.go
@@ -66,7 +66,7 @@ type DetachNodeRequest struct {
 
 type UpdateNodeStateRequest struct {
 	NodeId string `json:"node_id" path:"node_id" validate:"required"`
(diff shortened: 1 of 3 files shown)


salman-codewar (approved):

salman-codewar:
Thanks for fixes this 🙏

PR #618, opened 2023-11-08
Console app and BFF deployment fixes + Improvements

salman-codewar:
Issues covered: #619 

**Additional enhancements:**

- Added `getNodesLocation` & `getNodeLocation` queries
- Added `getStatMetrics` query
- Hooked `getNodesLocation` & `getStatMetrics` on home page


interfaces/console/Dockerfile
```diff
@@ -1,62 +1,27 @@
-# FROM nginx
-# COPY  build /usr/share/nginx/html
-# RUN rm /etc/nginx/conf.d/default.conf
-# COPY nginx/nginx.conf /etc/nginx/conf.d
-# EXPOSE 8080
-# CMD ["nginx", "-g", "daemon off;"]
+FROM node:18-alpine AS base
```
Brackleycassinga, line 1: Can you use node:alpine instead of node node:18 to reduce the sieze of the image?

interfaces/console/Dockerfile
```diff
@@ -1,62 +1,27 @@
...
-# EXPOSE 3000
-
-# ENV PORT 3000
+FROM base AS deps
+RUN apk add --no-cache libc6-compat
+WORKDIR /app
 
-# CMD ["yarn", "start"]
+COPY package.json yarn.lock* package-lock.json* pnpm-lock.yaml* ./
+RUN \
```
Brackleycassinga, line 8: Can you use multi stage build and maybe combine the run commands:
I think you can do something like this:
```
# Build stage
FROM node:alpine AS build
WORKDIR /app
COPY package*.json ./
RUN npm ci --only=production
COPY . .
RUN npm run build

# Production stage
FROM node:alpine
WORKDIR /app
COPY --from=build /app/package*.json ./
COPY --from=build /app/.next ./.next
COPY --from=build /app/public ./public
ENV NODE_ENV=production
CMD ["npm", "start"]
```

interfaces/console/pages/site_planning.tsx
```diff
@@ -168,13 +168,13 @@ const Page = () => {
     variables: {
       userId: _commonData.userId,
     },
-    onCompleted: (data) => {
+    onCompleted: (data: any) => {
```
Brackleycassinga, line 171: I think we have a type of this data https://github.com/ukama/ukama/blob/bf8d25017e2b9f8e5ac0f937589c2563f9ff44af/interfaces/console/generated/index.ts#L749C13-L749C18

systems/console-bff/Dockerfile
```diff
@@ -1,14 +1,15 @@
 FROM node
```
Brackleycassinga, line 1: Again we can use alpine image here .

interfaces/console/package.json
```diff
@@ -60,6 +60,7 @@
     "react-xarrows": "^2.0.2",
     "recoil": "^0.7.7",
     "recoil-persist": "^4.2.0",
+    "sharp": "^0.32.6",
```
Brackleycassinga, line 63: What is this lib for ?

interfaces/console/pages/_app.tsx
```diff
@@ -11,23 +11,43 @@ import { commonData, isDarkmode, snackbarMessage } from '@/app-recoil';
...
+import { useEffect } from 'react';
 import { RecoilRoot, useRecoilState, useRecoilValue } from 'recoil';
 import '../styles/global.css';
-
 const MainApp = dynamic(() => import('@/pages/_main_app'));
 const clientSideEmotionCache = createEmotionCache();
 const SNACKBAR_TIMEOUT = 5000;
 
+const getMetaInfo = async () => {
+  return await fetch('https://api.ipify.org/?format=json', {
```
Brackleycassinga, line 30: Can you move  this api url in constant ?

interfaces/console/pages/site_planning.tsx
```diff
@@ -187,7 +187,7 @@ const Page = () => {
   });
 
   const [addDraftCall, { loading: addDraftLoading }] = useAddDraftMutation({
-    onCompleted: (data) => {
+    onCompleted: (data: any) => {
```
Brackleycassinga, line 190: We have a type for this

interfaces/console/Dockerfile
Brackleycassinga, line 1: Can you use node:alpine instead of node node:18 to reduce the sieze of the image?

systems/console-bff/Dockerfile
```diff
@@ -1,14 +1,15 @@
...
```
Brackleycassinga, line 2: Again we can use alpine image here .

interfaces/console/package.json
salman-codewar: Next. js built-in image optimization requires sharp as a dependency
https://www.npmjs.com/package/sharp

interfaces/console/Dockerfile
salman-codewar: No it won't bring any size difference, "node:alpine" will use latest node image which is 21 and by mentioning "18-alpine" it will use 18 version.
We need to test app with latest version then we can use `node:alpine`

systems/console-bff/Dockerfile
salman-codewar: Good catch, Thanks

interfaces/console/pages/site_planning.tsx
salman-codewar: Addressed

interfaces/console/pages/site_planning.tsx
salman-codewar: addressed

interfaces/console/pages/_app.tsx
salman-codewar: Addressed

PR #626, opened 2023-12-22
common/rest: Move systems rest clients (resty wrappers) to common. Fixes #604

### LINKED ISSUE #604: Common REST client > Move systems rest clients (resty wrappers) to common
A lot  of times, services within a given system  need to interact with other services within other systems. This interaction is mostly done through REST calls to the destination system's API Gateway (gRpc is solely used for internal system communication).  These REST call are essentially performed by small wrappers around Resty client and because the wrapper to get a given service resource is almost always identical, regardless from where the REST call originates, we end up with some considerable amount of duplicated code.  This duplication occurs in various systems in general and in particular in API system and testing repository, which both heavily rely on REST call to other systems.

We need to abstract the wrapper for each system and place it into our common lib so that it can be reused by other systems, in order to avoid any kind of duplication.

delvatt:
This pull request fixes #604  . It adds the following changes:

- Move systems resty clients into their own packages inside common
- Update system's services to use systems clients from common


salman-codewar (approved):
Thanks for this PR man. Cleaned up lot of duplicated code 🔥 🚀

Brackleycassinga (approved):
Looks good to me (👍🏾 Great work !
Please fix the conflicts with main branch.

PR #627, opened 2024-01-08
Enhancement/network service fields

### LINKED ISSUE #645: Enhancement - Network Service Fields - Pull Request #627
Subject: Enhancement - Network Service Fields - Pull Request #627

## Description
This pull request, #627, aims to enhance the capabilities of the network service by introducing new fields such as city, language, currency, and country. The proposed additions are designed to improve the overall functionality of the network service, allowing for more granular searches and the configuration of networks with specific currencies and locations.

## Changes
- Added +100 / -185 lines across 18 files.

## Proposed Changes
The new fields introduced in this pull request will significantly enhance the capabilities of the network service. Users will be able to retrieve networks based on currency, and the ability to configure a network with a different currency or location will be valuable for various use cases.

## Checklist
- [ ] Currency, country and language fields added to network model
- [ ] Code has been reviewed
- [ ] Tests have been added or updated to cover the changes

Brackleycassinga:
This pull request adds new fields to improve the capabilities of the network service, including city, language, currency, and country. These additions will enable us to retrieve networks based on currency, as well as configure a network with a different currency or location. Fixes #645


systems/registry/docker-compose.yml
```diff
@@ -52,6 +52,9 @@ services:
    - NETWORK_SERVICE_PORT=9090
    - PUSHGATEWAY=http://${LOCAL_HOST_IP}:9099
    - ORGHOST=http://${LOCAL_HOST_IP}:8060
+   - COUNTRY=Democratic Democratic Republic of the Congo
+   - CURRENCY=USD
+   - LANGUAGE=EN
```
salman-codewar, line 57: We should use some code standards for country, currency and language to be consistent.

**Some suggestion**
For country we can use:
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3#Officially_assigned_code_elements

For currency we can use:
https://en.wikipedia.org/wiki/ISO_4217#Active_codes

For language code we can use:
https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes

systems/registry/api-gateway/pkg/rest/router.go
```diff
@@ -53,7 +53,7 @@ type Clients struct {
 
 type network interface {
 	AddNetwork(orgName, netName string, allowedCountries, allowedNetworks []string, budget, overdraft float64,
-		trafficPolicy uint32, paymentLinks bool) (*netpb.AddResponse, error)
+		trafficPolicy uint32, paymentLinks bool ) (*netpb.AddResponse, error)
```
salman-codewar, line 56: We remove this change

systems/registry/network/pkg/config.go
```diff
@@ -32,6 +32,9 @@ type Config struct {
 	Service          *uconf.Service
 	PushGateway      string `default:"http://localhost:9091"`
 	OrgName          string
+	Country          string `default:"Democratic Republic of the Congo"`
+	Language         string `default:"en"`
+	Currency         string `default:"usd"`
```
salman-codewar, line 37: I don't think we should not set default values at model level for these properties. Let's set them via config.

salman-codewar (changes requested):

systems/registry/network/pkg/config.go
Brackleycassinga: Okay

systems/registry/docker-compose.yml
Brackleycassinga: We will incorporate checks at the BFF level to invoke APIs for validating country, currency, and language, thereby enforcing the standards for these fields.

salman-codewar (approved):

PR #722, opened 2024-07-04
BFF deployment fixes

salman-codewar:
We were facing some issues with Bff deployment, trying to resolve these issues in this PR
- BFF docker compose added
- Config updates


Brackleycassinga (approved):
LGTM

PR #725, opened 2024-07-15
Sites page integration

Brackleycassinga:


interfaces/console/src/app/console/sites/[id]/page.tsx
```diff
@@ -7,47 +7,233 @@
...
+        type: 'error' as AlertColor,
+        show: true,
+      });
+    },
+  });
+  const handleSiteConfiguration = async (data: any) => {
+    const variables = {
+      access_id: data.access,
+      backhaul_id: data.backhaul,
+      install_date: getCurrentDateInISOFormat(),
```
Comment (salman-codewar) [inline interfaces/console/src/app/console/sites/[id]/page.tsx:124]: We are using `date-fns` package which can help us here. On the sites page i have replace it with 
`formatISO(new Date())` & formatISO is imported from `date-fns` like `import { formatISO } from 'date-fns';`

interfaces/console/src/components/SiteInfos/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+    return await fetch(
+      `https://nominatim.openstreetmap.org/reverse?format=json&lat=${lat || 37.7749}&lon=${lng || -122.4194}`,
+      {
+        cache: 'force-cache',
+      },
+    )
+      .then((res) => res.json())
+      .then((data) => data.display_name)
+      .catch(() => 'Location not found');
+  };
```
salman-codewar, line 29: Seems like we are using this method multiple places, Can we common it out. Create a `repo` directory under `src` and move it to there.

salman-codewar (approved):

Issue #718, opened 2024-06-24
Console BFF integration testing for subgraphs


salman-codewar:
To improve code quality & reliability, We need to implement e2e integration test for BFF subgraphs.
https://www.apollographql.com/docs/apollo-server/testing/testing/#end-to-end-testing

Lets try to achieve 70% coverage of integration tests.

PR #728, opened 2024-07-16
Console-bff Integration Tests

Asfand-Yar-Hassan:
- Integration tests for console-bff subgraps
[Issue#718](https://github.com/ukama/ukama/issues/718)


systems/console-bff/gateway/tests/invitation.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      email
+      expireAt
+      id
+      name
+      role
+      link
+      userId
+      status
+    }
+  }`;
```
salman-codewar, line 86: I think it will be nice if we move queries and mutations variable in a separate file. 
We can create `graphql.ts` file under tests and move all these mutation /queries in it and import it wherever needed.

systems/console-bff/gateway/tests/invitation.test.ts
```diff
@@ -0,0 +1,332 @@
...
+            redisClient.isOpen ? redisClient : null
+          );
+          return {
+            dataSources: {
+              dataSource: invitationAPi,
+            },
+            baseURL: baseURL.message,
+            headers: parsedHeaders,
+          };
+        })(),
```
salman-codewar, line 108: Just noticed this Context Value code is duplicated multiple time. We can define this in `beforeAll` and assign its value to a variable then use it everywhere. It'll reduce lot of code duplication.

salman-codewar (approved):

salman-codewar (approved):

salman-codewar:
Thanks @Asfand-Yar-Hassan  for this addition, Looks good 👍 💯

Issue #727, opened 2024-07-16
Subscriber auth relation with subscriber registry


salman-codewar:
- Make `email` required in subscriber model.
- Add Get subscriber by email request

Issue #726, opened 2024-07-15
Add package is not storing "amount" and "ownerId"


salman-codewar:
<img width="700" alt="Screenshot 2024-07-15 at 9 26 55 PM" src="https://github.com/user-attachments/assets/c8347875-c0a3-45a3-be50-0ffedb96cfd2">

PR #729, opened 2024-07-18
Add Package bug fix, Get subscriber by email added

salman-codewar:
Addressed issues #726 #727


builder/deploy_config.json
```diff
@@ -13,5 +13,5 @@
  },
  "key": "the-key-has-to-be-32-bytes-long!",
  "lago-api-key": "ad5ca9af-9d60-47e3-9b8c-43d354d5f6c7",
- "systems": "auth-services, services, init, nucleus, registry, dataplan, subscriber, notification, bff, console"
+ "systems": "auth-services, services, init, nucleus, registry, dataplan, subscriber"
```
Brackleycassinga, line 16: No need to commit this change since we need all systems to be there by default.

systems/common/pb/ukama/subscriber.proto
```diff
@@ -21,7 +20,7 @@ import "google/protobuf/wrappers.proto";
     string lastName = 3 [json_name = "last_name"];
     string subscriberId = 4 [(validator.field) = {uuid_ver: 4, string_not_empty: true}, json_name = "subscriber_id"];
     string networkId = 5 [(validator.field) = {uuid_ver: 4, string_not_empty: true}, json_name = "network_id"];
-    string email  = 6;
+    string email  = 6 [(validator.field) = { human_error:"must be an email format" regex: "^$|^[a-z0-9._%+\\-]+@[a-z0-9.\\-]+\\.[a-z]{2,4}$"}];
```
Brackleycassinga, line 23: You can implement email validation at the system level. We've decided to use all validations from common.

systems/data-plan/package/pkg/db/package_repo.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 type PackageRepo interface {
-	Add(_package *Package) error
+	Add(packageData *Package, packageRate *PackageRate) error
```
Brackleycassinga, line 21: Maybe we can just name it `dataPackage`. I'm not sure why we're adding a data package and then returning the rate instead of the package itself.

systems/subscriber/registry/pb/subscriber.proto
```diff
@@ -87,6 +87,18 @@ message GetSubscriberResponse{
 ukama.common.v1.Subscriber subscriber = 1;
 }
 
+// GetSubscriberByEmailRequest message is used to get a subscriber by subscriber email
+message GetSubscriberByEmailRequest{
+// email field is used to get subscriber by email
+string email = 1 [(validator.field) = { human_error:"must be an email format" regex: "^$|^[a-z0-9._%+\\-]+@[a-z0-9.\\-]+\\.[a-z]{2,4}$"}];
```
Brackleycassinga, line 93: Need to use validation from common.

systems/subscriber/registry/pkg/server/subscriber_test.go
```diff
@@ -152,6 +152,31 @@ func TestSubscriberServer_Get(t *testing.T) {
 		subRepo.AssertExpectations(t)
 	})
 
+	// t.Run("SubscriberByEmail", func(t *testing.T) {
```
Brackleycassinga, line 155: Why is this test commented out?

Brackleycassinga (changes requested):

systems/common/pb/ukama/subscriber.proto
salman-codewar: Please open an issue on it, This need to be address on all systems. We can address it in separate PR.

systems/data-plan/package/pkg/db/package_repo.go
salman-codewar: Here `packageRate` is an argument not a return type.

systems/subscriber/registry/pb/subscriber.proto
salman-codewar: Validator is already coming from common.

systems/subscriber/registry/pkg/server/subscriber_test.go
salman-codewar: oh by mistake.

systems/subscriber/registry/pb/subscriber.proto
Brackleycassinga: Then we don't need to have extra validation here it should just be string

Brackleycassinga (approved):

PR #773, opened 2024-10-01
Node service updates and fixes

salman-codewar:
PR addressed:

- Moved node connectivity and state enums to common
- Removed unnecessary rpc's


systems/common/ukama/node_state.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"database/sql/driver"
+	"strconv"
+	"strings"
+)
+
+type NodeState uint8
+
+const (
+	Undefined NodeState = iota
```
Brackleycassinga, line 20: We have changed these states to :
unknown, config, operational ,faulty

systems/registry/node/pkg/db/model.go
```diff
@@ -151,3 +62,20 @@ func (s *Site) BeforeSave(tx *gorm.DB) (err error) {
 
 	return nil
 }
+
+func ParseNodeState(s string) ukama.NodeState {
```
Brackleycassinga, line 66: This parsing func should also be part of common i guess.

systems/registry/node/pkg/config.go
```diff
@@ -62,6 +62,7 @@ func NewConfig(name string) *Config {
 		},
 		Service: uconf.LoadServiceHostConfig(name),
 		MsgClient: &uconf.MsgClient{
+			// Host:    "msgclient-registry:9095",
```
Brackleycassinga, line 65: This msg client host comment can be removed.

systems/registry/node/pkg/db/model.go
salman-codewar: addressed

systems/registry/node/pkg/config.go
salman-codewar: addressed

Brackleycassinga (approved):

PR #774, opened 2024-10-01
Update Event Configuration with Missing Site Event

Brackleycassinga:
- Added `siteAdd` and `siteUpdate` events to the common `event.config`.
- Updated the event-notification listener to handle these site-related events.


systems/notification/event-notify/pkg/server/event.go
```diff
@@ -378,6 +378,32 @@ func (es *EventToNotifyEventServer) EventNotification(ctx context.Context, e *ep
...
+		msg, err := epb.UnmarshalEventAddSite(e.Msg, c.Name)
+		if err != nil {
+			return nil, err
+		}
+		jmsg, err := json.Marshal(msg)
+		if err != nil {
+			log.Errorf("Failed to store raw message for %s to db. Error %+v", c.Name, err)
+		}
+
+		_ = es.ProcessEvent(&c, es.orgId, "", "", "", "", jmsg)
```
salman-codewar, line 392: Site event also have `NetworkId` which we can pass to `ProcessEvent` function.

salman-codewar (approved):

delvatt (approved):

PR #783, opened 2024-10-09
NodeState Events Config added to Common

Brackleycassinga:
Add NodeStateEventsConfig to manage node state events and their routing keys. It includes different node states, like create, assign, release, and online/offline , etc . Each event has a corresponding routing key to help with event handling. This change will make it easier to manage node transitions in node state srv.

Changes Made:

Added NodeStateEventId type and NodeStateEventsConfig structure.
Created constants for various node state events.
Defined routing keys for each event.
Mapped events to their config for better understanding and use.


Brackleycassinga:
This PR will be closed because the changes are already incorporated into the state machine branch.

PR #788, opened 2024-10-15
Notification system updates

salman-codewar:
- Added actionable notifications types
- Notifications resourceId bug fix
- Subscriber notification flow issue fixes


systems/common/events/event.go
```diff
@@ -274,15 +274,15 @@ var EventToEventConfig = map[EventId]EventConfig{
...
+		Type:        notif.TYPE_ACTIONABLE_INFO,
 	},
 	EventMeshNodeOffline: {
 		Key:         EventMeshNodeOffline,
 		Name:        "EventMeshNodeOffline",
 		Title:       "Mesh Node Offline",
 		Description: "Mesh Node Offline",
 		Scope:       notif.SCOPE_ORG,
-		Type:        TypeDefault,
+		Type:        notif.TYPE_ACTIONABLE_INFO,
```
Brackleycassinga, line 285: Why is "node offline" something we need to act on? I don't think it needs any action. It should just be TYPE INFO .

systems/common/notification/notification.go
```diff
@@ -35,11 +35,15 @@ func (l NotificationScope) Value() (driver.Value, error) {
 type NotificationType uint8
 
 const (
-	TYPE_INAVLID  NotificationType = 0
-	TYPE_INFO     NotificationType = 1
-	TYPE_WARNING  NotificationType = 2
-	TYPE_ERROR    NotificationType = 3
-	TYPE_CRITICAL NotificationType = 4
+	TYPE_INAVLID             NotificationType = 0
```
Brackleycassinga, line 38: There is a typo here it should be `INVALID`.

Brackleycassinga (approved):

systems/common/events/event.go
salman-codewar: This kind of notification supposed to lead to detail page. So in this case, if `ABC` node is offline that an actionable info which will lead to `ABC` notification detail page where use can perform curtain type of actions. like `restart`.

systems/common/notification/notification.go
salman-codewar: addressed

systems/common/rest/client/subscriber/subscriber.go
```diff
@@ -20,7 +20,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const SubscriberEndpoint = "/v1/subscribers"
+const SubscriberEndpoint = "/v1/subscriber"
```
delvatt, line 23: Why are we changing this?

delvatt (approved):

systems/common/rest/client/subscriber/subscriber.go
salman-codewar: it should be `subscriber`, that was a typo.

delvatt: Normally speaking, rest resources names inside endpoints should be plural.  We need to update this in subscriber system api-gtw

PR #793, opened 2024-10-24
Minor updates to Org and User service

salman-codewar:
- Added non-require `currency` property to Org
- User service removed phone number to be unique tag from model.


Brackleycassinga (approved):

delvatt (approved):
LGTM. Just explain why we are removing the unique tag for user phone.

PR #794, opened 2024-10-28
Subscriber system updates

salman-codewar:
This PR includes:
- Subscriber system updates related to subscriber app flow
- Sim allocation bug fixes
- Sim Package Activate/Deactivate timestamp fix
- Simpool sim allocation event listner
- Subscriber registry `firstname` & `lastname` update to `name`
- Subscriber auth session key updated


delvatt (approved):
LGTM. But this might break some rest client code in common that rely on some of the fields that we are removing. We need to update those as well, otherwise this will make this update incomplete.

vthakur7f (approved):

PR #808, opened 2024-11-07
Notify service msgClient routing keys and event cleanup

Brackleycassinga:
Added missing events:

- Node created
- Node online
- Node offline

Updated routing keys to: notify.node.store"


systems/node/notify/cmd/server/main.go
```diff
@@ -91,10 +90,6 @@ func runGrpcServer(gormdb sql.Db) {
 		srv := server.NewNotifyServer(serviceConfig.OrgName, db.NewNotificationRepo(gormdb), mbClient)
 		generated.RegisterNotifyServiceServer(s, srv)
 
-		eSrv := server.NewNotifyEventServer(serviceConfig.OrgName, db.NewNotificationRepo(gormdb), mbClient,
-			serviceConfig.MsgClient.ListenerRoutes)
-		egenerated.RegisterEventNotificationServiceServer(s, eSrv)
-
```
delvatt, line 97: This means we are removing the event server.

systems/node/notify/internal/config.go
```diff
@@ -27,23 +27,21 @@ type Config struct {
...
-			Host:    "msg-client-notification:9095",
 			Timeout: 5 * time.Second,
 			ListenerRoutes: []string{
-				"event.cloud.global.{{ .Org}}.nucleus.org.notification.sent",
-				"event.cloud.global.{{ .Org}}.nucleus.user.notification.sent",
-				"event.cloud.local.{{ .Org}}.registry.network.notification.sent",
-				"event.cloud.local.{{ .Org}}.registry.node.notification.sent",
+				"event.cloud.local.{{ .Org}}.messaging.mesh.node.online",
+				"event.cloud.local.{{ .Org}}.messaging.mesh.node.offline",
+				"event.cloud.local.{{ .Org}}.registry.node.node.create",
```
delvatt, line 43: But we are keeping some routes ?

systems/node/notify/internal/server/event.go
```diff
@@ -21,68 +23,176 @@ import (
 	mb "github.com/ukama/ukama/systems/common/msgBusServiceClient"
 	epb "github.com/ukama/ukama/systems/common/pb/gen/events"
 )
+ 
+ type NotifiyEventServer struct {
```
delvatt, line 27: And we still keep the Event server related code.

delvatt (commented):
So, now I am confused. Are we removing the Event server? If yes, then we should remove all related code including the routes. If no, then we should start the event server in `cmd/server/main`

systems/node/notify/internal/config.go
Brackleycassinga: The routes have been updated. Remember we had a chat with Vishal and decided to only have node related routes which were these four.

systems/node/notify/internal/server/event.go
Brackleycassinga: Yes, because we are going to handle node events as discussed.

Brackleycassinga: The event server code was not pushed my bad.it's fixed now.

systems/node/notify/internal/config.go
delvatt: If that''s the case, then you should start the event server. You are no long starting it, right now.

systems/node/notify/internal/server/event.go
```diff
@@ -21,68 +23,176 @@ import (
...
+	 orgName        string
+	 notifyRepo     db.NotificationRepo
+	 listenerRoutes map[string]struct{}
```
delvatt, line 30: You are not using this. Should be removed then.

systems/node/notify/internal/server/event.go
```diff
@@ -21,68 +23,176 @@ import (
...
+	 msgbus         mb.MsgBusServiceClient
+	 baseRoutingKey msgbus.RoutingKeyBuilder
+	 epb.UnimplementedEventNotificationServiceServer
+ }
+ 
+ func NewNotifyEventServer(orgName string, nRepo db.NotificationRepo, msgBus mb.MsgBusServiceClient, routes []string) *NotifiyEventServer {
+ 
+	 pRoutes := msgbus.PrepareRoutes(orgName, routes)
```
delvatt, line 38: You are not using this. Should be removed then.

systems/node/notify/internal/server/event.go
```diff
@@ -21,68 +23,176 @@ import (
...
+	 r := make(map[string]struct{}, len(routes))
+ 
+	 for _, route := range pRoutes {
```
delvatt, line 41: You are not using this. Should be removed then.

systems/node/notify/internal/server/event.go
```diff
@@ -21,68 +23,176 @@ import (
...
+		 r[route] = struct{}{}
+	 }
+ 
+	 return &NotifiyEventServer{
+		 orgName:        orgName,
+		 notifyRepo:     nRepo,
+		 listenerRoutes: r,
```
delvatt, line 48: You are not using this. Should be removed then.

systems/node/notify/internal/server/event.go
```diff
@@ -21,68 +23,176 @@ import (
...
+ type NotifiyEventServer struct {
+	 orgName        string
+	 notifyRepo     db.NotificationRepo
+	 listenerRoutes map[string]struct{}
+	 msgbus         mb.MsgBusServiceClient
+	 baseRoutingKey msgbus.RoutingKeyBuilder
+	 epb.UnimplementedEventNotificationServiceServer
+ }
+ 
+ func NewNotifyEventServer(orgName string, nRepo db.NotificationRepo, msgBus mb.MsgBusServiceClient, routes []string) *NotifiyEventServer {
```
delvatt, line 36: You are not using `routes []string`. Should be removed then.

systems/node/notify/internal/server/event_test.go
```diff
@@ -1,117 +0,0 @@
-/*
```
delvatt, line 1: We need the event_test.go.  Please add the tests for the event handlers.

systems/node/notify/internal/server/notify.go
```diff
@@ -119,7 +119,7 @@ func (n *NotifyServer) Delete(ctx context.Context, req *pb.GetRequest) (*pb.Dele
 		return nil, err
 	}
 
-	route := n.baseRoutingKey.SetAction("delete").SetObject("notification").MustBuild()
+	route := n.baseRoutingKey.SetAction("delete").SetObject("node").MustBuild()
```
delvatt, line 122: We are deleting a **notification**, not a "node"

systems/node/notify/internal/server/notify.go
```diff
@@ -215,7 +215,7 @@ func add(nodeId, severity, nType, serviceName string, details []byte, nStatus ui
 		return err
 	}
 
-	route := baseRoutingKey.SetAction("store").SetObject("notification").MustBuild()
+	route := baseRoutingKey.SetAction("store").SetObject("node").MustBuild()
```
delvatt, line 218: We are storing a **notification**, not a "node"

delvatt (commented):
Good overall, but we definitely needs the tests for the events server. Make sure you add those.

systems/node/notify/internal/server/notify.go
Brackleycassinga: We discussed last time and agree to change that to node as the key will have alreay notify which means we don't need notification again.node.notify.node.delete which is fine.

systems/node/notify/internal/server/notify.go
Brackleycassinga: We discussed last time and agree to change that to node as the key will have alreay notify which means we don't need notification again.node.notify.node.store .

systems/node/notify/internal/server/event.go
delvatt: This is still pending

systems/node/notify/internal/server/event.go
delvatt: This is still pending

systems/node/notify/internal/server/notify.go
delvatt: Not sure if I was part of this discussion since i can't remember this specific point, still, I disagree for this naming because of the following:

1. **It is misleading:**  A good event name should be meaningful and not misleading especially without knowing the context, eg the event name should tell you exactly what is happening even if you do not know the context.  For example, if you randomly ask a group developers who do not know anything about the ukama repository or code base and ask them what they thinks this event  `"event.cloud.local.{{ .Org}}.node.notify.node.store"`  is about, the vast majority will tell you this is an event about some node being stored, which is false, misleading and a clear indication that this naming is wrong. 
2. **It violates our routing key naming scheme:**  Our routing key specs looks like `event.cloud.<scope>.org-name>.<system-name>.<service-name>.<object>.<action>` . The object is the thing that is being manipulated by the action (or the thing that is causing the action) and here the object is notification (not node)  because this event is happening because we are storing one notification row inside the notification DB, eg. DB.Notifications.Store on gorm (Remember, the DB is called notification, because it holds a bunch of notifications, and we are storing one of these notification). It is not a DB of nodes from which we are storing one node item, so calling  `"event.cloud.local.{{ .Org}}.node.notify.node.store"` an event triggered by a DB.Notifications.Store gorm operation is telling a false story.  The only thing that can make `"event.cloud.local.{{ .Org}}.node.notify.node.store"` not wrong is if we change the name of DB from **Notifications** to **Nodes**. Then and only then, the naming will be consistent, even if it won't make any sense to call a DB of notifications as DB.Nodes
3. **You are conflicting Notify and Notifications**.  Notify is the name of the service. Notification is one group of ojects manipulated by the Notify service. That's why we have a Notification DB inside that service. But this service can have many other DBs. Remember these events are about what happening inside the notify service itself, so if an event inside the notify service is generated because we are storing one notification item inside the Notification DB from the notify service, then the associated event should be named `"event.cloud.local.{{ .Org}}.node.notify.notification.store"` just for the sake of being correct and consistent.
4. **And I can go on and on and on ...** on why this naming is violating our own design principles.  Now, I still don't remember agreeing on this naming in a previous discussion but we can yet have a discussion.

systems/node/notify/internal/server/notify.go
Brackleycassinga: Routing keys updated accordingly  @delvatt .

systems/node/notify/internal/server/notify.go
Brackleycassinga: Routing key updated @delvatt

systems/node/notify/internal/server/event.go
```diff
@@ -25,64 +27,158 @@ import (
...
+func (n *NotifiyEventServer) unmarshalNodeOnlineEvent(msg *anypb.Any) (*epb.NodeOnlineEvent, error) {
+	p := &epb.NodeOnlineEvent{}
+	err := anypb.UnmarshalTo(msg, p, proto.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true})
+	if err != nil {
+		log.Errorf("Failed to Unmarshal NodeOnline  message with : %+v. Error %s.", msg, err.Error())
+		return nil, err
+	}
+	return p, nil
+}
+func (n *NotifiyEventServer) unmarshalNodeOfflineEvent(msg *anypb.Any) (*epb.NodeOfflineEvent, error) {
```
salman-codewar, line 176: We have generated unmarshal methods in common try using those

salman-codewar (changes requested):

salman-codewar (approved):

delvatt (approved):

PR #815, opened 2024-11-14
Minor fixes after testing

salman-codewar:


delvatt (approved):
As long as @Brackleycassinga  can weight in the `ts`, `tsx` part, all the go part is godd. Good job!

Brackleycassinga (approved):

PR #829, opened 2025-01-02
Console Billing Page integration

Brackleycassinga:


systems/console-bff/common/data/correspondent.ts
```diff
@@ -0,0 +1,227 @@
+import { TCorrespondentsMap } from "../types";
```
salman-codewar, line 1: This is related to MobileMoney are we using it here?

systems/console-bff/payment/resolver/getCorrespondents.ts
```diff
@@ -0,0 +1,27 @@
+/*
```
salman-codewar, line 1: It's also related to mobile money, Double check if we really needed it here.

salman-codewar (approved):
Good job man.
Looks good, we might need some clean up & also make sure the billing notification is working.

systems/console-bff/payment/resolver/getCorrespondents.ts
Brackleycassinga: I included it because Kashif mentioned that we might support SaaS fee payments via mobile money in the future.

systems/console-bff/common/data/correspondent.ts
Brackleycassinga: Addressed

Brackleycassinga: addressed.

PR #907, opened 2025-03-04
Implement Dummy dsubscriber service

salman-codewar:


systems/common/rest/client/registry/node.go
```diff
@@ -36,10 +36,11 @@ type NodeStatusInfo struct {
 type NodeInfo struct {
 	Id        string         `json:"id,omitempty"`
 	Name      string         `json:"name,omitempty"`
-	OrgId     string         `json:"org_id,omitempty"`
 	Type      string         `json:"type,omitempty"`
 	Status    NodeStatusInfo `json:"status,omitempty"`
 	Site      NodeSiteInfo   `json:"site,omitempty"`
+	Latitude  float64        `json:"latitude,omitempty"`
```
Brackleycassinga, line 42: Why did you remove lat and lng here ?

testing/services/dummy/dsubscriber/README.md
```diff
@@ -0,0 +1,16 @@
+# Site Registry
```
Brackleycassinga, line 1: Can we remove this or update it? It's an old Markdown document for the registry site.

Brackleycassinga (changes requested):
LGTM (👍🏾 .

testing/services/dummy/dsubscriber/clients/agent.go
```diff
@@ -0,0 +1,76 @@
+/*
```
delvatt, line 1: Wouldn't it make more sense to move ukama agent CDR client under common/rest/client so that it could be use by any other service that might need i?

delvatt (approved):
LGTM with some few comments: 

- Sim manager was made before we decided to use strings for dates in proto , instead of timestampp.pb.  For the dates previously in timestamp, we will update when we'll fix code coverage, but for new dates like package_end_date  we should use strings for dates as we previously agreed. 
-  Also lets make sure that ukama agent CDR client is not duplicate.
- Also let's make sure we are fixing all failing workflows.

systems/common/pb/events/simmanager.proto
```diff
@@ -43,6 +43,7 @@ message EventSimAllocation {
     bool isPhysical = 11 [json_name = "is_physical"];
     string packageId = 12 [(validator.field) = {uuid_ver: 4, string_not_empty: true}, json_name = "package_id"];
     uint32 trafficPolicy = 13 [json_name = "traffic_policy"];
+    google.protobuf.Timestamp packageEndDate = 14 [json_name = "end_date"];
```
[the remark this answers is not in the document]
salman-codewar: It's not being updated everywhere, In common events it mostly `timestamp`.
Here's a ref from `main`
https://github.com/ukama/ukama/blob/961c76fcf66858847bd92bba4ec313a8c5bac28a/systems/common/pb/events/simmanager.proto#L54

We might need to open an issue and make it consistent everywhere.

testing/services/dummy/dsubscriber/clients/agent.go
salman-codewar: It's node-gateway integration. I'm sure if any other system would want to push cdr's to node-gateway.

PR #1003, opened 2025-05-13
Usage:  update error handling. Fixes #997

### LINKED ISSUE #997: Handle get usage request for new subscriber
We faced a case where we were getting error when new subscriber (with sim allocation) was created and console was trying to get the usage for that subscriber. Sim manager was throwing error:

Image

We are getting this error because for new user there is no CDR in agent to calculate the usage. But even in that case rather throwing the error we can send 0 usage. This issue is to address the behaviour to getUsage api in sim manger in case of new subscriber.

delvatt:
This PR fixes #997  by providing the following: 

- Update error handling on ASR's query usage
- Update error handling on CDR's query usage

Fixes #997


salman-codewar (approved):
LGTM.
Just a note, `common/mocks` changes increase PR size significantly. It'll be great if we avoid these version changes in mocks.

Brackleycassinga (approved):
LGTM 👍🏾

PR #1030, opened 2025-06-02
Fix Subscriber Deletion Flow, Active SIM Count Metrics, and Usage Reset Bug

Brackleycassinga:


systems/common/ukama/subscriber_status.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+
+type SubscriberStatus uint8
+
+const (
+	SubscriberStatusUnknown        SubscriberStatus = iota 
+	SubscriberStatusActive                               
+	SubscriberStatusPendingDeletion   
+	SubscriberStatusInactive
```
delvatt, line 24: Why not simply have `SubscriberStatusDeleted` instead. **SubscriberStatusInactive**  here is misleading because anything inactive can become active and vice versa, which matches more a pausing behavior.  Here we are implementing a deletion behavior, and that state is definitive. So it makes more sense to call the associated status `SubscriberStatusDeleted`

systems/common/pb/events/asr.proto
```diff
@@ -27,3 +28,15 @@ message AsrInactivated{
 message AsrUpdated {
     Subscriber Subscriber = 1;
 }
+
+message SimCleanupResult {
+  string simId = 1;
```
delvatt, line 33: `simId` should not be leaked into ASR. Under agent systems, we only identify sims with ICCID or IMSI. Agents should not have any knowledge of simIds.

systems/common/pb/events/asr.proto
```diff
@@ -27,3 +28,15 @@ message AsrInactivated{
...
+  string iccid = 2;
+  bool success = 3;
+}
+
+message EventSimAsrCleanupCompleted {
+  string subscriberId = 1 [(validator.field) = {uuid_ver: 4, string_not_empty: true}, json_name = "subscriber_id"];
```
delvatt, line 39: Same here, I am not sure that subscriberId should be leaked into ASR. Under Agents, "The ICCID/Imsi is the subscriber". There should be no representation of any physical or moral subscriber on ASR.

systems/notification/event-notify/pkg/server/event.go
```diff
@@ -488,8 +488,11 @@ func (es *EventToNotifyEventServer) EventNotification(ctx context.Context, e *ep
 		if err != nil {
 			log.Errorf("Failed to store raw message for %s to db. Error %+v", c.Name, err)
 		}
+		dynamicConfig := c
```
delvatt, line 491: Why making a copy of c, instead of updating it directly?

systems/subscriber/registry/cmd/server/main.go
```diff
@@ -58,16 +61,37 @@ func initConfig() {
...
+func initDb() sql.Db {
+    log.Infof("Initializing Database")
+    d := sql.NewDb(serviceConfig.DB, serviceConfig.DebugMode)
+    err := d.Init(&db.Subscriber{})
+
+    if err != nil {
+        log.Fatalf("Database initialization failed. Error: %v", err)
+    }
+    
+    if err := d.GetGormDb().Exec("CREATE UNIQUE INDEX IF NOT EXISTS idx_subscribers_active_email ON subscribers (email) WHERE deleted_at IS NULL").Error; err != nil {
```
delvatt, line 74: NO! We shouldn't be doing that.  If we already have a DB without those unique indexes, then we should delete that DB and start we a new  one, having Gorm applying all the constraints defined in the model from scratch. 
Then later, when ready, we'll have separate processes for applying migrations on existing DBs.

systems/subscriber/registry/cmd/server/main.go
```diff
@@ -95,11 +119,14 @@ func runGrpcServer(gormdb sql.Db) {
 
 	simMClient := client.NewSimManagerClientProvider(serviceConfig.SimManagerHost)
 
-	srv := server.NewSubscriberServer(serviceConfig.OrgName, db.NewSubscriberRepo(gormdb), mbClient, simMClient, serviceConfig.OrgId, orgClient, networkClient)
-
+	srv := server.NewSubscriberServer(serviceConfig.OrgName, db.NewSubscriberRepo(gormdb), mbClient, simMClient, serviceConfig.OrgId, orgClient, networkClient, serviceConfig)
```
delvatt, line 122: Either you pass individual serviceConfig.Values or you pass the whole serviceConfig (which I personnaly don't like, but anyway ... ) but you do not do both.

systems/subscriber/registry/pkg/config.go
```diff
@@ -41,6 +48,18 @@ func NewConfig(name string) *Config {
 		Service: uconf.LoadServiceHostConfig(name),
 		MsgClient: &uconf.MsgClient{
 			Timeout: 5 * time.Second,
+			ListenerRoutes: []string{
+				"event.cloud.local.{{ .Org}}.subscriber.simmanager.sims.deletion_completed",
```
delvatt, line 52: We shoud follow routing keys convention: no underscore  "sims.delete" seems fine.

systems/subscriber/registry/pkg/db/model.go
```diff
@@ -11,18 +11,22 @@ package db
...
+	"github.com/ukama/ukama/systems/common/ukama"
 	uuid "github.com/ukama/ukama/systems/common/uuid"
 )
  
  type Subscriber struct {
 	 SubscriberId          uuid.UUID `gorm:"primaryKey;type:uuid"`
 	 Name             string    `gorm:"size:255"`
 	 NetworkId             uuid.UUID `gorm:"type:uuid;index"`
-	 Email                 string    `gorm:"size:255;unique;not null;index"`
+	 Email                 string    `gorm:"size:255;not null;index"`
```
delvatt, line 22: The unique index constraint combo the deletedAt should be added here. Not in ` cmd/server/main.go`

systems/subscriber/registry/pkg/db/model.go
```diff
@@ -11,18 +11,22 @@ package db
...
 	 NetworkId             uuid.UUID `gorm:"type:uuid;index"`
-	 Email                 string    `gorm:"size:255;unique;not null;index"`
+	 Email                 string    `gorm:"size:255;not null;index"` 
 	 PhoneNumber           string    `gorm:"size:15"`
 	 Gender                string    `gorm:"size:255"`
 	 DOB                   string
 	 ProofOfIdentification string `gorm:"size:255"`
+	 SubscriberStatus 			  ukama.SubscriberStatus 
+	 DeletionRetryCount    int                      `gorm:"default:0"`
+     DeletionLastAttempt   *time.Time
```
delvatt, line 29: why are we using a pointer?

systems/subscriber/registry/pkg/db/subscriber_repo.go
```diff
@@ -22,6 +25,9 @@ type SubscriberRepo interface {
 	Update(subscriberId uuid.UUID, sub Subscriber) error
 	GetByNetwork(networkId uuid.UUID) ([]Subscriber, error)
 	ListSubscribers() ([]Subscriber, error)
+	FindPendingDeletionBefore(threshold time.Time) ([]Subscriber, error)
+    MarkAsPendingDeletion(subscriberId uuid.UUID) error
+    IncrementDeletionRetry(subscriberId uuid.UUID) error
```
delvatt, line 30: Not a big deal, but `FindPendingDeletionBefore(threshold time.Time) ([]Subscriber, error)`  and `GetByNetwork(networkId uuid.UUID)` could have been merged into a **proper ListSubscriber()**, if one existed
and both `MarkAsPendingDeletion(subscriberId uuid.UUID) error` and `IncrementDeletionRetry(subscriberId uuid.UUID) error` could be merged into **Update(subscriberId uuid.UUID, sub Subscriber) error**

systems/subscriber/registry/pkg/db/subscriber_repo.go
```diff
@@ -116,3 +122,52 @@ func (s *subscriberRepo) GetByNetwork(networkId uuid.UUID) ([]Subscriber, error)
 	}
 	return subscribers, nil
 }
+
+func (s *subscriberRepo) FindPendingDeletionBefore(threshold time.Time) ([]Subscriber, error) {
```
delvatt, line 126: ListSubcribers()

systems/subscriber/registry/pkg/db/subscriber_repo.go
```diff
@@ -116,3 +122,52 @@ func (s *subscriberRepo) GetByNetwork(networkId uuid.UUID) ([]Subscriber, error)
...
+    var subscribers []Subscriber
+    result := s.Db.GetGormDb().Where("subscriber_status = ? AND deletion_last_attempt < ?", 
+        ukama.SubscriberStatusPendingDeletion, threshold).Find(&subscribers)
+    return subscribers, result.Error
+}
+
+func (s *subscriberRepo) MarkAsPendingDeletion(subscriberId uuid.UUID) error {
```
delvatt, line 133: Update()

systems/subscriber/registry/pkg/db/subscriber_repo.go
```diff
@@ -116,3 +122,52 @@ func (s *subscriberRepo) GetByNetwork(networkId uuid.UUID) ([]Subscriber, error)
...
+    }
+    if result.RowsAffected == 0 {
+        return gorm.ErrRecordNotFound
+    }
+    
+    return nil
+}
+
+
+func (r *subscriberRepo) IncrementDeletionRetry(subscriberId uuid.UUID) error {
```
delvatt, line 155: Update()

systems/subscriber/registry/pkg/server/event.go
```diff
@@ -0,0 +1,272 @@
...
+		orgName: orgName,
+		s:       s,
+	}
+}
+
+func (es *RegistryEventServer) EventNotification(ctx context.Context, e *epb.Event) (*epb.EventResponse, error) {
+	log.Infof("Received a message with Routing key %s and Message %+v", e.RoutingKey, e.Msg)
+
+	switch e.RoutingKey {
+	case msgbus.PrepareRoute(es.orgName, "event.cloud.local.{{ .Org}}.subscriber.simmanager.sims.deletion_completed"):
```
delvatt, line 45: no underscore

systems/subscriber/registry/pkg/server/event.go
```diff
@@ -0,0 +1,272 @@
...
+		if err != nil {
+			return nil, err
+		}
+	default:
+		log.Errorf("No handler routing key %s", e.RoutingKey)
+	}
+
+	return &epb.EventResponse{}, nil
+}
+func (es *RegistryEventServer) handleSimDeactivation(ctx context.Context, event *epb.EventSimDeactivation) error {
```
delvatt, line 98: No, that is wrong.

systems/subscriber/registry/pkg/server/event.go
```diff
@@ -0,0 +1,272 @@
...
+	err = es.s.subscriberRepo.Update(subscriberId, subscriber)
+	if err != nil {
+		log.Errorf("Failed to update subscriber %s status to inactive: %v", subscriberId, err)
+		return err
+	}
+	
+	log.Infof("Updated subscriber %s status to inactive due to SIM deactivation", subscriberId)
+	return nil
+}
+func (es *RegistryEventServer) handleSimActivation(ctx context.Context, event *epb.EventSimActivation) error {
```
delvatt, line 120: No, That is wrong

systems/subscriber/registry/pkg/server/event.go
```diff
@@ -0,0 +1,272 @@
...
+		err = es.handleEventCompleteSubscriberDeletion(e.RoutingKey, msg)
+		if err != nil {
+			return nil, err
+		}
+	case msgbus.PrepareRoute(es.orgName, "event.cloud.local.{{ .Org}}.ukamaagent.asr.activesubscriber.delete"):
+		msg, err := es.unmarshalAsrInactivated(e.Msg)
+		if err != nil {
+			return nil, err
+		}
+		err = es.handleSubscriberDeactivated(ctx, msg)
```
delvatt, line 60: I don't believe this should be systematic all the time.

systems/subscriber/registry/pkg/server/event.go
```diff
@@ -0,0 +1,272 @@
...
+
+func (es *RegistryEventServer) unmarshalSimDeactivation(msg *anypb.Any) (*epb.EventSimDeactivation, error) {
+	p := &epb.EventSimDeactivation{}
+	err := anypb.UnmarshalTo(msg, p, proto.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true})
+	if err != nil {
+		log.Errorf("Failed to unmarshal EventSimDeactivation message: %+v. Error: %s", msg, err.Error())
+		return nil, err
+	}
+	return p, nil
+}
```
delvatt, line 272: All this need to be reworked.

systems/subscriber/registry/pkg/server/subscriber.go
```diff
@@ -370,39 +381,74 @@ func (s *SubcriberServer) Update(ctx context.Context, req *pb.UpdateSubscriberRe
...
+    deletionEvent := &epb.EventSubscriberDeletionInitiated{
+        SubscriberId: subscriber.SubscriberId.String(),
+        Name:         subscriber.Name,
+        Sims:         simDetails,
+    }
+
+    err = s.PublishEventMessage(route, deletionEvent)
+    if err != nil {
+        log.Errorf("Failed to publish deletion initiation event: %v", err)
+        return nil, status.Errorf(codes.Internal, "Failed to initiate deletion: %v", err)
```
delvatt, line 444: wrong error message

systems/subscriber/sim-manager/pkg/config.go
```diff
@@ -61,8 +61,10 @@ func NewConfig(name string) *Config {
 				"event.cloud.local.{{ .Org}}.subscriber.simmanager.sim.allocate",
 				"event.cloud.local.{{ .Org}}.payments.processor.payment.success",
 				"event.cloud.local.{{ .Org}}.ukamaagent.asr.activesubscriber.delete",
+				"event.cloud.local.{{ .Org}}.subscriber.registry.subscriber.delete",
 				"event.cloud.local.{{ .Org}}.ukamaagent.cdr.cdr.create",
 				"event.cloud.local.{{ .Org}}.operator.cdr.cdr.create",
+				"event.cloud.local.{{ .Org}}.ukamaagent.asr.subscriber.asr_cleanup_completed",
```
delvatt, line 67: no underscore

systems/subscriber/sim-manager/pkg/server/sim_manager.go
```diff
@@ -558,6 +558,144 @@ func (s *SimManagerServer) GetSimsByNetwork(ctx context.Context, req *pb.GetSims
...
+				IsActive:  false,
+				AsExpired: true,
+			}
+
+			err = s.packageRepo.Update(packageToTerminate, func(pckg *sims.Package, tx *gorm.DB) error {
+				packageToTerminate.EndDate = time.Now().UTC()
+				return nil
+			})
+			if err != nil {
+				return fmt.Errorf("failed to terminate active package %s: %w", pkg.Id, err)
```
delvatt, line 686: But are the packages deleted?

systems/subscriber/sim-manager/pkg/server/sim_manager.go
```diff
@@ -589,6 +727,34 @@ func (s *SimManagerServer) TerminateSim(ctx context.Context, req *pb.TerminateSi
...
 	}
 
+	otherSims, err := s.simRepo.List("", "", sim.SubscriberId.String(), "", ukama.SimTypeUnknown, ukama.SimStatusUnknown, 0, false, 0, false)
+	if err != nil {
+		log.Errorf("Failed to check for other SIMs for subscriber %s: %v", sim.SubscriberId, err)
+		return nil, grpc.SqlErrorToGrpc(err, "sims")
+	}
+
+	if len(otherSims) == 1 {
+    log.Infof("SIM %s is the only SIM for subscriber %s, triggering full subscriber deletion cascade",
```
delvatt, line 737: Are we sure about that? and even if yes, this check/trigger should be in subscriber registry, not sim-manager

systems/ukama-agent/asr/pkg/config.go
```diff
@@ -54,6 +54,7 @@ func NewConfig(name string) *Config {
 			Timeout: 5 * time.Second,
 			ListenerRoutes: []string{
 				"event.cloud.local.*.ukamaagent.cdr.cdr.create",
+				"event.cloud.local.{{ .Org}}.subscriber.registry.subscriber.deletion_initiated",
```
delvatt, line 57: no underscore?

systems/ukama-agent/asr/pkg/server/event.go
```diff
@@ -51,13 +54,126 @@ func (l *AsrEventServer) EventNotification(ctx context.Context, e *epb.Event) (*
...
+			return err
+		}
+		log.Infof("Successfully published ASR cleanup completion for subscriber: %s", subscriberId)
+	} else {
+		log.Warnf("Message bus client not available, cannot publish ASR cleanup completion")
+	}
+
+	return nil
+}
+func (l *AsrEventServer) deleteAsrRecordByIccid(iccid string) error {
```
delvatt, line 149: Duplicate of ASR.Inactivate?

delvatt (changes requested):
We need to rework some of those pieces.

systems/common/ukama/subscriber_status.go
Brackleycassinga: We use SubscriberStatusInactive because on the console, when the last SIM for a subscriber becomes inactive, we want to reflect the subscriber's status as inactive. This status can later change back to active if the last (or only) SIM is reactivated.

Using SubscriberStatusDeleted would imply a permanent and irreversible deletion, which isn’t accurate in this case—since the subscriber can become active again if their SIM is reactivated.

systems/subscriber/registry/cmd/server/main.go
Brackleycassinga: Last time, we agreed that deleted subscribers should be allowed to re-register using the same email address. However, if we rely solely on GORM’s constraints, this won’t work—because the unique constraint still applies even to soft-deleted records. This means that attempting to add a subscriber with the same email (as a previously deleted one) will fail.

systems/common/pb/events/asr.proto
Brackleycassinga: The simId in SimCleanupResult, you're right—it shouldn't be exposed in ASR. Agent systems should only identify SIMs by iccid or imsi. I'll make the necessary changes to align with that.

systems/common/pb/events/asr.proto
Brackleycassinga: I initially included the subscriberId because I was thinking ahead—other services might listen to this in the future. But yes, I can update it to only include iccid for now.

systems/notification/event-notify/pkg/server/event.go
Brackleycassinga: Addressed.

systems/subscriber/registry/cmd/server/main.go
Brackleycassinga: addressed

systems/subscriber/registry/pkg/config.go
Brackleycassinga: Addressed.

systems/subscriber/registry/pkg/db/model.go
Brackleycassinga: Addressed.

systems/subscriber/registry/pkg/db/model.go
Brackleycassinga: I used a Pointer here because :

- New subscribers - haven't had any deletion attempts yet → should be NULL
- Active subscribers - no deletion attempts → should be NULL
- Subscribers being deleted - have deletion attempts → has a timestamp value

systems/subscriber/registry/pkg/db/subscriber_repo.go
Brackleycassinga: Yeah, that makes sense. We just need a listSubscribers endpoint with support for filters.

systems/subscriber/registry/pkg/db/subscriber_repo.go
Brackleycassinga: Addressed.

systems/subscriber/registry/pkg/db/subscriber_repo.go
Brackleycassinga: Addressed

systems/subscriber/registry/pkg/db/subscriber_repo.go
Brackleycassinga: Addressed.

systems/subscriber/registry/pkg/server/event.go
Brackleycassinga: Addressed

systems/ukama-agent/asr/pkg/server/event.go
Brackleycassinga: Addressed.

systems/ukama-agent/asr/pkg/config.go
Brackleycassinga: Addressed

systems/subscriber/sim-manager/pkg/server/sim_manager.go
Brackleycassinga: Need to be discussed.

PR #1087, opened 2025-08-04
Bootstrap service aded & Node gateway updated

salman-codewar:


systems/common/rest/client/factory/factory.go
```diff
@@ -0,0 +1,76 @@
...
+func (s *nodeFactoryClient) Get(id string) (*NodeFactoryInfo, error) {
+	log.Debugf("Getting node by id from factory: %v", id)
+
+	nodeFactory := NodeFactoryInfo{}
+
+	resp, err := s.R.Get(s.u.String() + FactoryEndpoint + "/node/" + id)
+	if err != nil {
+		log.Errorf("Get node failure. error: %s", err.Error())
+
+		return nil, fmt.Errorf("getComponent failure: %w", err)
```
delvatt: `_getNodeFatctory_ failure` instead of **getComponent failure.**

systems/init/bootstrap/pkg/server/bootstrap.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, err
+	}
+
+	var ip string
+	var certificate string
+
+	if node.OrgName != "" {
+		msgSystem, err := lookupSvc.GetSystemForOrg(ctx, &lpb.GetSystemRequest{
+			OrgName:    node.OrgName,
+			SystemName: "messaging",
```
delvatt: Maybe put this hardcoded `"messaging"` string  into a constant, so that it will be easy to update in the future.

systems/init/node-gateway/pkg/client/bootstrap.go
```diff
@@ -0,0 +1,67 @@
...
+	client  pb.BootstrapServiceClient
+	timeout time.Duration
+	host    string
+}
+
+func NewBootstrap(bootstrapHost string, timeout time.Duration) *Bootstrap {
+
+	conn, err := grpc.NewClient(bootstrapHost, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Fatalf("did not connect: %v", err)
```
delvatt: We need to update this log message `"did not connect"` everywhere we can because this is not insightful at all. If we are ever in a context where we do not have the Logging Prefix, then we will have no idea where this log is coming from.  So maybe something like `"Failed to connect to Bootstrap Service:  %v"` should be better.

systems/init/node-gateway/pkg/client/bootstrap.go
```diff
@@ -0,0 +1,67 @@
...
+func NewBootstrapFromClient(mClient pb.BootstrapServiceClient) *Bootstrap {
+	return &Bootstrap{
+		host:    "localhost",
+		timeout: 1 * time.Second,
+		conn:    nil,
+		client:  mClient,
+	}
+}
+
+func (r *Bootstrap) Close() {
```
delvatt, line 58: The following might be even better and avoids the connection nil pointer exception error.
```
func (r *Bootstrap) Close() {
	if r.conn != nil {
		if err := r.conn.Close(); err != nil {
			log.Warnf("Failed to gracefully close Bootstrap Service connection: %v", err)
		}
	}
}
```

PR #1093, opened 2025-09-18
Lint and build fixes on multiple services

salman-codewar:


systems/init/lookup/internal/server/lookup.go
```diff
@@ -103,7 +102,7 @@ func (l *LookupServer) AddOrg(ctx context.Context, req *pb.AddOrgRequest) (*pb.A
 }
 
 func (l *LookupServer) UpdateOrg(ctx context.Context, req *pb.UpdateOrgRequest) (*pb.UpdateOrgResponse, error) {
-	logrus.Infof("Updating Organization %s", req.OrgName)
+	log.Infof("Updating Organization %s", req.OrgName)
 	req.GetIp()
```
Brackleycassinga, line 106: This can  `req.GetIp() `be removed as it may cause issues if left unused.

systems/init/lookup/internal/server/lookup.go
```diff
@@ -242,7 +241,7 @@ func (l *LookupServer) GetNode(ctx context.Context, req *pb.GetNodeRequest) (*pb
 }
 
 func (l *LookupServer) GetNodeForOrg(ctx context.Context, req *pb.GetNodeForOrgRequest) (*pb.GetNodeResponse, error) {
-	logrus.Infof("Get node %s for org %s.", req.GetNodeId(), req.GetOrgName())
+	log.Infof("Get node %s for org %s.", req.GetNodeId(), req.GetOrgName())
 
 	_, err := l.orgRepo.GetByName(req.OrgName)
```
Brackleycassinga, line 246: What’s the purpose of this  	`_, err := l.orgRepo.GetByName(req.OrgName) `if the return value isn’t being used?

testing/services/dummy/api-gateway/pkg/client/dsubscriber.go
```diff
@@ -51,7 +51,11 @@ func NewHealthFromClient(mClient pb.DsubscriberServiceClient) *Dsubscriber {
 }
 
 func (r *Dsubscriber) Close() {
-	r.conn.Close()
+	if r.conn != nil {
+		if err := r.conn.Close(); err != nil {
+			logrus.Errorf("failed to close connection: %v", err)
```
Brackleycassinga: Use `log` alias instead of `logrus`.

systems/init/lookup/internal/server/lookup.go
salman-codewar: This change is not related to this PR. Since i didn't develop it that's why idk the side effects of removing it.

systems/init/lookup/internal/server/lookup.go
salman-codewar: This change is not related to this PR.

delvatt (approved):
LGTM!

PR #1139, opened 2025-12-02
Inventory components update for node flow

salman-codewar:
> [!IMPORTANT]
> This PR introduces a scheduler for node synchronization, adds new gRPC methods for component management, and updates mock implementations for testing.
> 
>   - **Scheduler**:
>     - Introduces `ComponentScheduler` in `scheduler.go` for scheduling node sync jobs.
>     - Adds `StartScheduler` and `StopScheduler` methods in `component.go` to control the scheduler.
>   - **gRPC Methods**:
>     - Adds `StartScheduler`, `StopScheduler`, and `Verify` RPC methods in `component.proto`.
>     - Implements `Verify` method in `component.go` to check component existence.
>   - **Mocks**:
>     - Updates mock files for `ComponentServiceClient` and `ComponentServiceServer` to include new methods.
>   - **Database**:
>     - Adds `Verify` method in `component_repo.go` to check component by part number.
>   - **Testing**:
>     - Adds tests for new scheduler and component verification functionalities in `component_test.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ukama%2Fukama&utm_source=github&utm_medium=referral)<sup> for ed0d5586f519347da7a1efd959658d21c6683b99. You can [customize](https://app.ellipsis.dev/ukama/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/systems/inventory/component/pkg/config.go
+++ b/systems/inventory/component/pkg/config.go
@@ -44,7 +44,7 @@ type NodeComponentDetails struct {
 	Managed       string                  `default:"true"`
 	Category      ukama.ComponentCategory `default:"ACCESS"`
 	Manufacturer  string                  `default:"Ukama Inc"`
-	Inventory     string                  `default:"ukma-access"`
+	Inventory     string                  `default:"ukama-access"`
 	DatasheetURL  string                  `default:"http://www.ukama.com/datasheet"`
 }
 


systems/inventory/component/pkg/db/component_repo.go
```diff
@@ -36,6 +36,15 @@ func NewComponentRepo(db sql.Db) ComponentRepo {
 	}
 }
 
+func (c *componentRepo) Verify(partNumber string) (*Component, error) {
```
delvatt, line 39: So basically and unless I am wrong,  `Verify` is just another `GetComponentByPartNumber` if that is the case, then, why not simply use **List and filtering by PartNumber** ?

Also, from my understanding, PartNumber is only guarantee to be unique only within the realm of a manufacturer.  If this is true, then we are at risk with possible (even if not probable) collisions.

salman-codewar: We already have List endpoint. Verify endpoint is just for `mesh` which doesn't care about other components and just want to know that part number is valid or not. We wanted to make it lightweight that's why its a separate endpoint.

This inventory stuff is not related to manufacturing.  It's for org which bring there own part's like backhaul, solar, batteries.

PR #1210, opened 2026-03-06
Node software service update

salman-codewar:
- Minor hub gateway updates to upload file artifacts from swagger page
- Node software service updated, Which keep track of each app version per node. Also record if new update is available
- Node api gateway updates related to software service

### DIFF
--- a/systems/common/validation/version.go
+++ b/systems/common/validation/version.go
@@ -9,19 +9,15 @@
 package validation
 
 import (
-	"net/http"
+	"errors"
 
 	"github.com/coreos/go-semver/semver"
-	"github.com/ukama/ukama/systems/common/rest"
 )
 
 func ParseVersion(version string) (*semver.Version, error) {
 	v, err := semver.NewVersion(version)
 	if err != nil {
-		return nil, rest.HttpError{
-			HttpCode: http.StatusBadRequest,
-			Message:  "Invalid version format. Refer to https://semver.org/ for more information",
-		}
+		return nil, errors.New("Invalid version format. Refer to https://semver.org/ for more information")
 	}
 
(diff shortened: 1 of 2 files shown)


Brackleycassinga (approved):

systems/common/validation/version.go
```diff
@@ -0,0 +1,42 @@
...
+	"net/http"
+
+	"github.com/coreos/go-semver/semver"
+	"github.com/ukama/ukama/systems/common/rest"
+)
+
+func ParseVersion(version string) (*semver.Version, error) {
+	v, err := semver.NewVersion(version)
+	if err != nil {
+		return nil, rest.HttpError{
```
delvatt: I believe a better approach for this common library would be, in case of failure,  to just return a standard error and let the client of this library wraps the standard error and returns the proper specific error depending on the client type (http client will return a httpError, grpc client will return a grpcError,  CLI client will return a stdErr, etc ...)

systems/node/api-gateway/pkg/client/softwarer.go
```diff
@@ -16,6 +16,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
```
delvatt, line 16: `systems/node/api-gateway/pkg/client/softwarer.go`
Is the name `softwarer` legit? just wanna be sure it is not a typo

salman-codewar: yea that's being used everywhere for disabling.

systems/common/validation/version.go
salman-codewar: addressed

Issue #1323, opened 2026-05-19
Add PCRF + OVS/br0 support to virtual tower node with starter-managed EPC emulator for UE-to-media E2E traffic testing.


markhor:
- [x] 1. Update virtualnode builder:  tower only: include pcrf.d, epc-emu.d, OVS/br0 setup.

- [x] 2. Update tower manifest:  starter starts pcrf.d and epc-emu.d (boot).
 
- [x] 3. Add OVS setup script/service:  create br0, configure OpenFlow, default drop UE subnet.
 
- [x] 4. Add epc-emu.d:  REST attach/detach API for UE containers.  Calls PCRF session create/delete.  Provides UE traffic interface into OVS. Implementation.
 
- [x] 5. Wire virtual tower networking:  UE side -> epc-emu.d -> br0 -> media side.
 
- [x] 6. Create UE container runner:  spin N-UEs with IMSI/IP.  attach, generate traffic, detach.
 
- [x] 7. Create media container:   iperf3/HTTP/ping target.
 
- [x] 8. Add E2E validation script:    verify attach creates OVS flows,  verify traffic passes.   verify counters/CDRs update in real backend.  verify detach removes flows and traffic stops.
 

```
UE containers  <-->  virtual tower node  <-->  media container
                         |
                         |
                     backend
```

PR #1328, opened 2026-05-20
Add starter-compatible PCRF for virtual tower-node OVS flow. (#1323)

### LINKED ISSUE #1327: integrate pcrf as starter app
Currently pcrf is not available as starter app. Missing service port, missing EPs (/v1/ping, /v1/version, etc.) - needed by starter; and update pcrf config to work with the init-network ovs bridges

markhor:
This PR adds PCRF support for the virtual tower-node flow and aligns PCRF with the starter-managed app contract.

Key changes:
- Added starter-compatible root endpoints:
  - GET /v1/ping
  - GET /v1/version
  - GET /v1/status
- Removed the /pcrf route prefix so PCRF APIs now live directly under /v1.
- Updated PCRF to discover local service ports from /etc/services instead of hardcoded config:
  - pcrf
  - init-network
  - noded
- Updated PCRF startup to read OVS/runtime network state from init-network /v1/status.
- Removed duplicated PCRF-side OVS bridge configuration; init-network is now the source of truth for:
  - bridge name
  - OVS management socket
  - OpenFlow version
  - UE CIDR
- Updated PCRF DB location to /ukama/apps/db/pcrf.db and ensured the DB directory is created at startup.
- Added datapath status reporting so /v1/status exposes OVS connection state, management socket, and active UE datapath count.
- Reworked OVS datapath initialization to return errors instead of fatal exits, allowing PCRF startup/status to fail cleanly.
- Fixed the active-session store query bug caused by `WHERE WHERE state = 1`.
- Kept init-network, PCRF, and EPC boundaries clean:
  - init-network owns OVS/br0/default flows
  - PCRF owns per-UE policy flows/meters
  - EPC-emu will own attach/detach and user-plane emulation

Validation checklist:
- Confirm /etc/services contains pcrf, init-network, and noded entries.
- Start virtual tower node with init-network first in boot.
- Verify init-network is ready:
  - curl localhost:<init-network-port>/v1/status
  - ovs-vsctl show
  - ovs-ofctl -O OpenFlow15 dump-flows br0
- Start PCRF and verify:
  - curl localhost:<pcrf-port>/v1/ping returns OK
  - curl localhost:<pcrf-port>/v1/version returns generated build version
  - curl localhost:<pcrf-port>/v1/status shows datapath connected=true
- Confirm PCRF creates/uses DB at /ukama/apps/db/pcrf.db.
- Confirm existing PCRF APIs are reachable under /v1/... with no /pcrf prefix.


delvatt (approved):

delvatt:
@markhor  Also make sure you create an issue for the refactoring, so that we do not forget

PR #1366, opened 2026-06-04
Operation lock implementation 

### LINKED ISSUE #1372: Operation lock/monitor mechanism
Implement locking mechanism and update node domain system accordingly.

some potential rules to follow:

- Every mutating operation MUST be represented as an Action.
- Every Action MUST declare the resource or resources it intends to mutate.
- The backend MUST acquire a lock for each mutable resource before executing the Action.
- A resource MUST NOT have more than one active write Action at a time.
- Read-only operations MUST NOT require a lock.
- The backend MUST reject conflicting write Actions while a resource is locked.
- The backend SHOULD return the active Action when rejecting a conflicting  Action.
- An Action MUST NOT be marked succeeded when the command is merely accepted.
- An Action MUST be marked succeeded only after the expected final state is verified.
- Every Action MUST have a timeout.
- Every lock MUST have an expiry.
- The backend MUST release the lock when the Action reaches a terminal state.
- The backend SHOULD extend the lock only while the Action is actively making progress.
- Clients SHOULD use idempotency keys for write requests.
- The UI MUST treat the backend Action state as the source of truth.
- The UI SHOULD disable conflicting write controls while a resource is locked.
- The UI MAY continue to allow read-only views during active Actions.
- Force unlock MUST be restricted to privileged users and MUST be audited. (future version but it might be needed to break locks/debug)

Brackleycassinga:


markhor (changes requested):
per slack

PR #1406, opened 2026-06-17
Dataplan sys > Package updates

### LINKED ISSUE #1400: DataPlan > Packages add unique name validation

### LINKED ISSUE #1405: Data plan > needs to be per network
We need to have an option to assign (or during creation) valid data plan to a particular network. "ALL" should also be intentional and not by default.

salman-codewar:
- Make package name unique
- Add an ability to create package by network


delvatt (commented):
Not sure why we are adding a chain of endpoints for checking name availability, when we've already added a unique constraint on the name. I believe this should be enough to handle name uniqueness (just like how do for emails or unique IDs)

salman-codewar:
> Not sure why we are adding a chain of endpoints for checking name availability, when we've already added a unique constraint on the name. I believe this should be enough to handle name uniqueness (just like how do for emails or unique IDs)

To avoid the request failure, and let user know realtime if name is unique or not

delvatt (approved):
LGTM

PR #1424, opened 2026-06-23
Package, Invitation & Member delete flow minor fixes

### LINKED ISSUE #1423: Packages, Invitation & Member svc's delete flow fixes
- [x] Soft delete package with cascade effect
- [x] Soft delete invitation
- [x] Soft delete member, but member must be deactivate first before delete
- [x] Enable soft delete for Nodes
- [x] Enable soft delete for Sites
- [x] Enable soft delete for Networks

salman-codewar:


Brackleycassinga (approved):

systems/data-plan/package/pkg/server/package_test.go
```diff
@@ -53,8 +53,8 @@ const (
 	TestSimType               = "ukama_data"
 	TestPackageTypePrepaid    = "prepaid"
 	TestPackageTypePostpaid   = "postpaid"
-	TestDataUnitMB            = "MegaBytes"
-	TestDataUnitGB            = "GigaBytes"
```
delvatt, line 57:
```
	TestDataUnitMB            = "MegaBytes"
	TestDataUnitGB            = "GigaBytes"
```
Even this will still work.

systems/registry/network/pkg/server/network.go
```diff
@@ -10,6 +10,7 @@ package server
```
delvatt, line 10: I did not see much updates into the DeleteNetwork RPC.  Shouldn't we prevent deleting a network if we still have registered subscribers + sims under that network? 
Or put it in another way, shouldn't there be a validation step, checking that there is no longer any subscriber + sim registered under that network before we are allowed to delete it?

systems/registry/site/pkg/server/site.go
```diff
@@ -249,6 +249,37 @@ func (s *SiteServer) Update(ctx context.Context, req *pb.UpdateRequest) (*pb.Upd
 	}, nil
```
delvatt, line 249: Same reflexion here. Shouldn't we prevent deleting a site if we still have registered node under that site?

or put it in another way, shouldn't we have a validation step, that  check that there is no longer any node registered under this site before we allow to delete the site.

delvatt (approved):
I have left out some reflections as comments into few places. Make sure you have a look and provide your thoughts before you move on.

PR #1489, opened 2026-07-16
feat(sim-manager): record cash payment sale on SIM allocation

### LINKED ISSUE #1490: Record package sale as a payment on SIM allocation
When a SIM is allocated with a first data package, record that package sale as
a cash payment in the payments system — the same way top-ups are recorded — so
allocations appear in the payment ledger and have a downloadable receipt.

Allocation behaviour must not change: the SIM, the first package, the
sim.allocate event, ASR activation, billing, and the allocation email all stay
exactly as they are today.

## Background
Top-ups already record a cash payment (payment.success → sim-manager adds the
package). We initially tried to make allocation fully payment-driven too (create
the package only via payment.success), but that broke provisioning: the
sim.allocate event went out with an empty PackageId, and ukama-agent/asr's
activate() does uuid.FromString(packageId) on it (hard failure) and stores it as
Asr.SimPackageId, which the package-expiry flow later uses to promote the next
package. The package row must exist and be referenced at allocation time.

## Approach (Fix 1)
Keep the package creation synchronous at allocation (so PackageId is real and
everything downstream is unchanged), and additionally record the sale as a cash
payment. Mark that payment `provisioned` in metadata; sim-manager's
payment.success handler skips adding the package for provisioned payments, so
there's no duplicate package and no second email. The payment is still a valid
record — billing/report marks it paid and the receipt resolves via item_id + sim.

## Changes
- `common/rest/client/dataplan`: expose package `Currency`.
- `common/rest/client/payments`: new `PaymentClient` (POST /v1/payments) + mock + test.
- `sim-manager`: inject the payment client; `AllocateSim` records a provisioned
  cash payment (amount/currency from the data plan, metadata.sim + provisioned);
  `payment.success` handler skips `addPackageForSim` for provisioned payments.
- `payments/processor`: test asserting the cash `payment.success` event forwards
  metadata unchanged (the contract sim-manager depends on).

## Acceptance criteria
- [x] Allocating a SIM records a completed cash payment for the package.
- [x] The first package is still created inline; `sim.allocate` carries a real
      `PackageId`; ASR activation and billing are unchanged.
- [x] No duplicate package and no second email from the payment flow.
- [x] The allocation payment is retrievable and shows on the subscriber's
      package receipt (item_id + metadata.sim).
- [x] A failed payment record logs and does not fail allocation.

salman-codewar:


systems/subscriber/sim-manager/pkg/server/event.go
```diff
@@ -159,11 +159,20 @@ func (es *SimManagerEventServer) handleProcessorPaymentSuccessEvent(key string,
 		return fmt.Errorf("failed to Unmarshal payment metadata as map[string]string: %w", err)
 	}
 
-	simId, ok := metadata["sim"]
+	simId, ok := metadata[metadataSimKey]
```
delvatt, line 162: This previous "sim" metadata key might be in used in other systems (billing???). Make sure it is updated everywhere as needed

delvatt (commented):
This is a good start, but we should consider this as a WIP because there are other systems like ukama-agent that are drastically impacted by this change and need to be updated. If we merge this for now, these systems are going to be broken.

systems/subscriber/sim-manager/pkg/server/event.go
salman-codewar: thats just a const, actual key is `sim`

systems/subscriber/sim-manager/pkg/server/event.go
```diff
@@ -159,11 +159,20 @@ func (es *SimManagerEventServer) handleProcessorPaymentSuccessEvent(key string,
...
 	if !ok {
 		return fmt.Errorf("missing sim metadata for successful package payment: %s", msg.ItemId)
 	}
 
+	// Payments recorded by sim allocation already provisioned the package
+	// (added inline in AllocateSim), so skip adding it again here. The payment
+	// is still a valid sale record; only the package-add is suppressed.
+	if metadata[metadataProvisionedKey] == metadataProvisionedYes {
```
delvatt, line 170: I guess this is the part i've missed. If we are skipping `addPackageForSim` when already provisioned and we are sure that the `provisioned` metadata goes from sim manager to payment and back to sim manager, then this is fine.

systems/subscriber/sim-manager/pkg/server/event.go
delvatt: Sorry, my bad. my mind added the `""` around `metadataSimKey`. It is Friday 😛

delvatt (approved):
LGTM. 
But we need to be aware that when s.paymentClient is nil we do nothing. At least we should log that a payment record from a sim allocation is going to be skipped

PR #1535, opened 2026-07-24
feat(report): print payment receipts for completed one-off payments

### LINKED ISSUE #1487: Report System: Add support for payment receipts printing
We have invoice printing for subscription based models (org subscriptions, postpaid packages)
We need payments receipts printing for one off payments (prepaid packages, etc)

These belong to the report system, since we want those to be back-end generated and S3 bucket archive-able.

Brackleycassinga:
Screenshot 2026-07-31 at 12 58 11 PM


systems/billing/report/pkg/db/report_repo.go
```diff
@@ -20,6 +20,7 @@ import (
 type ReportRepo interface {
```
delvatt, line 20: I am not sure why we are making changes in billing/report. Let's go  through that on Monday so that I can understand if there is anything I am mssing.

Brackleycassinga: @delvatt I added `GetByTransactionId` to avoid creating duplicate receipts. The payment "update" event can arrive more than once for the same payment, so before we create a receipt we check whether we've already made one for that payment. We store the payment ID in the transaction_id column and look it up by that. The existing Get only finds a report by its own ID, which we don't have yet at that point, so we
  needed a lookup by transaction ID instead.

delvatt: But report under billing are data payloads only for billing system, they have nothing to do with payments system (which is another system that have nothing to do with billing system)
billing/report =>(produces)  invoice report payloads  => (which are sent) to report system for invoice file printings
payments/processor => (produces) one off payments payloads => (which are sent) to report system for payments receipts file printings

payments updates with status completed come from payment system (not from billing/report) with payment ID tha I believe we attribute to the report_id in the report system.

Again we can discuss and review together on monday

delvatt (approved):
LGTM!

salman-codewar (approved):

PR #1580, opened 2026-08-07
Node state ready race

Brackleycassinga:


Brackleycassinga:
This branch had health-node-notification merged into it so I could deploy and test on dev — the node chart's gRPC probes need that health service or the image crashloops. That merge pulled 95 files into the diff, most are unrelated.

PR #1582, opened 2026-08-07
Site-controller: restore desired service state after a tower restart

Brackleycassinga:
…rts drift


Brackleycassinga:
State model has changed so this is not relevant anymore .

PR #1613, opened 2026-08-20
Fix network node stale connectivity

### LINKED ISSUE #1585: Network node list reports stale connectivity after node disconnect
After a node disconnects, GetNode(nodeId) reports the node as Offline, but the network-scoped node list used by sitesView.nodeCounts still reports the same node as Online. This causes the Console site summary to show online=3, offline=0 even though one node is already offline.
`FAIL       site_node_counts_equals: matched=0/1 total=3 online=3 offline=0 expected={total=3 online=2 offline=1}
`

Problem details:
- Single-node lookup reflects the connectivity transition correctly.
- Network node lookup for the same network returns inconsistent/stale connectivity for the same node.
- This causes Console site node counts and degraded-site visibility to be incorrect.

**Scenraio:**
`scenarios/p0/console/sites/tc-021-one-node-unavailable-site-degraded.yaml`

Brackleycassinga:


delvatt (approved):
LGTM

PR #1619, opened 2026-08-21
Registry node > Fixed mismatch cause between node and site lat,lng 

salman-codewar:


salman-codewar:
Closing in favour of PR #1625

PR #1627, opened 2026-08-25
Reject node actions for offline nodes

### LINKED ISSUE #1621: Offline update requests are accepted but not run when the node reconnects
We should either not accept such request (preferable) or if we do then replay when online.

**Scenarios:**

```
p0/software-update/tc-055-software-update-controller-offline-catches-up.yaml
p0/software-update/tc-063-software-update-controller-partial-rollout.yaml
```

Brackleycassinga:


delvatt (approved):
LGTM

markhor (approved):

PR #1636, opened 2026-08-31
Keep a replaced instance from deleting the live registration and implemented reconcile behaviour

salman-codewar:


salman-codewar:
Closing in favour of https://github.com/ukama/ukama/pull/1637

PR #1711, opened 2026-09-18
Push cdr fixes. Fixes #1710

### LINKED ISSUE #1710: PCRF: PushCdr silently ignores non-2xx responses, causing permanent CDR-delivery failures
In` tc-010-two-sites-small` (and potentially any scenario with a terminated session), a UE's CDR sometimes never reaches the UKAMA AGENT CDR service at all — the usage_sample check times out with actual=0B even though the underlying data transfer completed successfully on the wire. Confirmed intermittent: reproduced on 2 separate runs during this investigation (usage=1/2 and usage=0/2), with the affected UE(s) varying — not tied to a specific site or session position.

delvatt:
This PR fixes #1710  and closes #1710 by providing the following:

- Enhance PushCdr goroutine to run safely
- Add a proper recover mechanism in case the PushCdr goroutine panics
- Fix swallowed http resty  error response
- Misc


Voyagerroc-Lab (approved):
Reviewed with assistance from OpenAI Codex. PushCdr now returns an error for non-2xx responses after the request error path, and the publisher isolates panics per handler so a failure in one CDR handler does not terminate the periodic goroutine or skip the other handler. The tests cover both non-2xx propagation and recovery boundaries.

